PR #13, opened 2021-08-31
Native bindings generated with abigen.

aguschirico:


connectors/rsk.go
```diff
@@ -151,183 +121,151 @@ func (rsk *RSK) GasPrice() (*big.Int, error) {
...
+	sizeInt, err := strconv.Atoi(results.String())
 	if err != nil {
-		return 0, fmt.Errorf("error parsing value from getFederationSize: %v", err)
+		return 0, fmt.Errorf("error converting federation size to int. error: %v", err)
 	}
 
 	return sizeInt, nil
 }
 
 func (rsk *RSK) GetFedPublicKeyOfType(index int) (string, error) {
```
javicid, line 192: Shouldn't be named OfType if it does not take the type

javicid (changes requested):

connectors/rsk.go
aguschirico: done.

javicid (approved):

PR #14, opened 2021-09-01
Added BTC functionality

javicid:


connectors/bitcoin.go
```diff
@@ -4,29 +4,155 @@ import (
...
+
+func (btc *BTC) AddAddressWatcher(address string, interval time.Duration, w AddressWatcher) error {
+	btcAddr, err := btcutil.DecodeAddress(address, &btc.params)
+	if err != nil {
+		return fmt.Errorf("error decoding address %v: %v", address, err)
+	}
+
+	err = btc.c.ImportAddressRescan(address, "", false)
+	if err != nil {
+		return fmt.Errorf("error importing address %v: %v", address, err)
```
aguschirico, line 65: shouldn't the original error message be enough here?

aguschirico (approved):

connectors/bitcoin.go
javicid: I am not sure, because sometimes the error returned by the call does not give context on what exactly failed. For example, it could say "method not found", and then you wouldn't know which is the method that is not found.

PR #17, opened 2021-09-02
Fixes and improvements to make the acceptQuote to work

aguschirico:


connectors/bitcoin.go
```diff
@@ -174,10 +174,10 @@ func (btc *BTC) GetDerivedBitcoinAddress(userBtcRefundAddr []byte, lbcAddress []
 	return addressScriptHash.EncodeAddress(), nil
 }
 
-func DecodeBTCAddress(address string) ([]byte, error) {
+func GetBytesFromBtcAddress(address string) ([]byte, error) {
```
javicid, line 177: I think that Decode is a better name consistent with other packages like hex.

javicid (changes requested):

javicid (approved):

connectors/bitcoin.go
aguschirico: Renamed.

PR #18, opened 2021-09-07
fix missing LP btc address in parseQuote

juampiq6:
I think the liquidity provider btc address was not being parsed when creating the struct **LiquidityBridgeContractQuote** that was then used to call the LBC function **hashQuote**.


aguschirico (approved):

javicid:
Closing because this has been included in #19

PR #23, opened 2021-09-17
Adds logic to retry peg-in if after required confirmations occurred the bridge still didn't acknowledge it.

aguschirico:


connectors/rsk.go
```diff
@@ -293,6 +294,10 @@ func (rsk *RSK) RegisterPegIn(opt *bind.TransactOpts, q bindings.LiquidityBridge
 	return rsk.lbc.RegisterPegIn(opt, q, signature, btcRawTrx, partialMerkleTree, height)
 }
 
+func (rsk *RSK) RegisterPegInWithoutTx(q bindings.LiquidityBridgeContractQuote, signature []byte, tx []byte, pmt []byte, height *big.Int, txHash string) error {
+	return rsk.lbc.RegisterPegInWithoutTx(q, signature, tx, pmt, height, txHash)
+}
+
```
javicid, line 300: This could be the same function as RegisterPegIn with an argument to decide whether to mine the transaction (using LBCCaller or LBC depending on the case).

http/watcher.go
```diff
@@ -115,10 +114,20 @@ func (w *BTCAddressWatcher) performRegisterPegIn(txHash string) error {
 	if err != nil {
 		return err
 	}
-	_, err = w.rsk.RegisterPegIn(opt, q, signature, rawTx, pmt, big.NewInt(int64(bh)))
+	err = w.rsk.RegisterPegInWithoutTx(q, signature, rawTx, pmt, big.NewInt(bh), txHash)
+	if err != nil {
+		if strings.Contains(err.Error(), "Failed to validate BTC transaction") {
```
javicid, line 119: Let's change the LBC so that the error code is in the error message, and look for the code instead of the whole text.

javicid (changes requested):

connectors/bindings/lbc.go
```diff
@@ -538,6 +538,19 @@ func (_LBC *LBCSession) RegisterPegIn(quote LiquidityBridgeContractQuote, signat
...
+func (l *LBCTransactor) RegisterPegInWithoutTx(q LiquidityBridgeContractQuote, signature []byte, tx []byte, pmt []byte, height *big.Int, txHash string) error {
+	opts := &bind.CallOpts{}
+	var res []interface{}
+	err := l.contract.Call(opts, &res, "registerPegIn", q, signature, tx, pmt, height)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
```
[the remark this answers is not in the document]
aguschirico: done.

connectors/rsk.go
aguschirico: I'd rather leaving it as it is for 2 main reasons:
1) making both signatures to match would imply 2 option params to be passed in or passing an useless transactopts to the function that uses the caller (and builds a callopts on the fly).
2) you can deduce what the function does based on its name, and looking it from the code that uses them it's easier to see what's intended.

http/watcher.go
aguschirico: ok that will be done afterwards right?

connectors/rsk.go
```diff
@@ -103,6 +105,7 @@ func (rsk *RSK) Connect(endpoint string) error {
 	if err != nil {
 		return err
 	}
+	rsk.lbcCaller = &bindings.LBCCallerRaw{Contract: &rsk.lbc.LBCCaller}
```
javicid, line 108: Why not create this when you need it instead of storing it as a class variable?

connectors/rsk.go
```diff
@@ -391,6 +394,15 @@ func (rsk *RSK) RegisterPegIn(opt *bind.TransactOpts, q bindings.LiquidityBridge
 	return rsk.lbc.RegisterPegIn(opt, q, signature, btcRawTrx, partialMerkleTree, height)
 }
 
+func (rsk *RSK) RegisterPegInWithoutTx(q bindings.LiquidityBridgeContractQuote, signature []byte, tx []byte, pmt []byte, height *big.Int) error {
```
javicid, line 397: Make parameter names the same as in RegisterPegIn for consistency (or the other way around).

http/watcher.go
```diff
@@ -74,11 +73,11 @@ func (w *BTCAddressWatcher) performCallForUser() error {
 		From:     q.LiquidityProviderRskAddress,
 		Signer:   w.lp.SignTx,
 	}
-
 	_, err = w.rsk.CallForUser(opt, q)
 	if err != nil {
 		return err
 	}
+	w.calledForUser = true
```
javicid, line 80: Here we could use GetTxStatus to wait for the tx being mined.

http/watcher.go
```diff
@@ -117,10 +116,15 @@ func (w *BTCAddressWatcher) performRegisterPegIn(txHash string) error {
...
 		return err
 	}
-	_, err = w.rsk.RegisterPegIn(opt, q, signature, rawTx, pmt, big.NewInt(int64(bh)))
+	err = w.rsk.RegisterPegInWithoutTx(q, signature, rawTx, pmt, big.NewInt(bh))
+	if err != nil {
+		if strings.Contains(err.Error(), "Failed to validate BTC transaction") {
+			return nil // allow retrying in case the bridge didn't acknowledge all required confirmations have occurred
+		}
+	}
+	_, err = w.rsk.RegisterPegIn(opt, q, signature, rawTx, pmt, big.NewInt(bh))
```
javicid, line 125: same here

javicid (approved):

http/watcher.go
aguschirico: I did in the same way the other usages do. Please lemme know if this what you meant.

connectors/rsk.go
aguschirico: done. I like this way more so I changed all to be the same as here.

connectors/rsk.go
aguschirico: Eitherway is fine as it's just a pointer to the contract. I did it originally this way cuz I thought it could be easier to access this in case of need. Moved it to a local variable.

http/watcher.go
aguschirico: done.

http/watcher.go
aguschirico: Exported getTxStatus and checking the status in watcher.

http/watcher.go
```diff
@@ -74,11 +76,19 @@ func (w *BTCAddressWatcher) performCallForUser() error {
...
+	tx, err := w.rsk.CallForUser(opt, q)
 	if err != nil {
 		return err
 	}
+	ctx, cancel := context.WithTimeout(context.Background(), connectors.EthTimeout)
+	defer cancel()
+	s, err := w.rsk.GetTxStatus(ctx, tx)
+	if err != nil {
+		return nil
+	}
```
javicid, line 88: If err != nil we should probably call GetStatus again, but not retry the transaction because it could have been mined anyways.

http/watcher.go
```diff
@@ -44,15 +45,16 @@ func (w *BTCAddressWatcher) OnNewConfirmation(txHash string, confirmations int64
...
 		}
-		w.calledForUser = true
 	}
 
 	if w.calledForUser && confirmations >= w.rsk.GetRequiredBridgeConfirmations() {
 		err := w.performRegisterPegIn(txHash)
 		if err != nil {
 			log.Errorf("error calling registerPegIn. value: %v. error: %v", txHash, err)
+			close(w.done)
 			return
```
javicid, line 58: These two lines are not needed

connectors/rsk.go
```diff
@@ -30,7 +30,7 @@ const (
 	rpcSleep   time.Duration = 2 * time.Second
 	rpcTimeout time.Duration = 5 * time.Second
 	ethSleep   time.Duration = 5 * time.Second
-	ethTimeout time.Duration = 5 * time.Minute
+	EthTimeout time.Duration = 5 * time.Minute
```
javicid, line 33: I am not sure about making this public since the idea of taking a context in getStatus is to be able to define a timeout outside of the connectos package. Using this value is conceptually the same as not taking a context as a parameter.

http/watcher.go
```diff
@@ -44,15 +45,16 @@ func (w *BTCAddressWatcher) OnNewConfirmation(txHash string, confirmations int64
 		err := w.performCallForUser()
 		if err != nil {
 			log.Errorf("error calling callForUser. value: %v. error: %v", txHash, err)
+			close(w.done)
 			return
 		}
-		w.calledForUser = true
```
javicid, line 49: Isn't this needed?

http/watcher.go
```diff
@@ -117,10 +127,25 @@ func (w *BTCAddressWatcher) performRegisterPegIn(txHash string) error {
...
+	tx, err := w.rsk.RegisterPegIn(opt, q, signature, rawTx, pmt, big.NewInt(bh))
 	if err != nil {
 		return err
 	}
+	ctx, cancel := context.WithTimeout(context.Background(), connectors.EthTimeout)
+	defer cancel()
+	s, err := w.rsk.GetTxStatus(ctx, tx)
+	if err != nil {
+		return nil
+	}
```
javicid, line 145: Same as in performCallForUser

javicid (changes requested):

http/watcher.go
```diff
@@ -74,11 +75,19 @@ func (w *BTCAddressWatcher) performCallForUser() error {
...
 	}
+	ctx, cancel := context.WithTimeout(context.Background(), time.Hour*8760) // timeout is a year
+	defer cancel()
+	s, err := w.rsk.GetTxStatus(ctx, tx)
+	if err != nil {
+		return err
+	}
+	if !s {
+		return fmt.Errorf("transaction failed. hash: %v", tx.Hash())
+	}
```
javicid: These can be in the same if statement

http/watcher.go
```diff
@@ -117,10 +126,25 @@ func (w *BTCAddressWatcher) performRegisterPegIn(txHash string) error {
...
```
javicid: These too

connectors/rsk.go
aguschirico: reverted as it's not necessary anymore.

http/watcher.go
aguschirico: it is. Accidentally removed it.

http/watcher.go
aguschirico: fixed.

http/watcher.go
aguschirico: I increased the timeout so we wait virtually forever for the getStatus to return something.

http/watcher.go
aguschirico: same as above, timeout increased so the tx gvies an actual result most likely.

javicid (approved):

PR #25, opened 2021-09-17
[Devops] Dockerization + Devops flow

rafaiovlabs:
* In order to work this needs secrets set up on the repo.


config.json
```diff
@@ -0,0 +1,48 @@
+{
```
aguschirico, line 1: this file shouldn't be in the repo. The configs here are secrets that should be managed outside the repo

rafaiovlabs: Probably the best way to address this is to define which data is sensitive but not erase the config..as its is needed for the app to run, both locally on other environments one possible solution is try: https://www.iotshaman.com/blog/content/application-secrets-in-nodejs-production-environments

PR #29, opened 2021-09-23
Add balance control for the LP

aguschirico:


http/server.go
```diff
@@ -216,18 +216,19 @@ func (s *Server) acceptQuoteHandler(w http.ResponseWriter, r *http.Request) {
 	}
 
 	p := getProviderByAddress(s.providers, quote.LPRSKAddr)
-	signB, err := p.SignHash(hashBytes)
+	cfuCost, err := s.rsk.EstimateGas(quote.ContractAddr, quote.Value, []byte(quote.Data))
```
javicid, line 219: This should be (quote.gasLimit + CFUExtraGas) * gasPrice + q.value

http/watcher.go
```diff
@@ -140,6 +130,17 @@ func (w *BTCAddressWatcher) performRegisterPegIn(txHash string) error {
...
+		return err
+	}
+	hb, err := hex.DecodeString(h)
+	if err != nil {
+		return err
+	}
+	err = w.lp.RefundLiquidity(hb)
+	if err != nil {
+		return fmt.Errorf("failed to refund to liquidity provider. quote: %v. error: %v", h, err)
+	}
```
javicid, line 144: I would put this outside this function. In the same place where close(w.done) is called.

javicid (changes requested):

http/watcher.go
aguschirico: done.

http/server.go
aguschirico: fixed.

javicid (approved):

PR #30, opened 2021-09-24
Fix retry peg-in flow

aguschirico:


connectors/rsk.go
```diff
@@ -386,11 +386,41 @@ func (rsk *RSK) GetLBCAddress() string {
...
+			break
+		}
+		time.Sleep(rpcSleep)
+	}
+	if tx == nil {
+		return nil, fmt.Errorf("error calling callForUser: %v", err)
+	}
+	if err != nil {
+		return nil, err
+	}
```
javicid, line 403: Is it possible for err to be != nil and t == nil?

javicid (changes requested):

javicid (approved):

javicid (approved):

javicid (approved):

PR #35, opened 2021-11-30
Added bitcoind connection check

Vovchyk:
- added check if a bitcoind node is connected on server start
- allowed gas price to be zero [here](https://github.com/rsksmart/liquidity-provider-server/pull/35/files#diff-4ab681cca396d706fb5744645ffd3c8a5b62661e6906797fb7236136fc451fc9R246)

### DIFF
--- a/connectors/bitcoin.go
+++ b/connectors/bitcoin.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"encoding/binary"
 	"fmt"
+	"github.com/btcsuite/btcd/btcjson"
 	"time"
 
 	"github.com/btcsuite/btcd/chaincfg"
@@ -64,6 +65,8 @@ func NewBTC(network string, fedInfo FedInfo) (*BTC, error) {
 		btc.params = chaincfg.MainNetParams
 	case "testnet":
 		btc.params = chaincfg.TestNet3Params
+	case "regtest":
+		btc.params = chaincfg.RegressionNetParams
 	default:
 		return nil, fmt.Errorf("invalid network name: %v", network)
 	}
@@ -84,6 +87,22 @@ func (btc *BTC) Connect(endpoint string, username string, password string) error
 	if err != nil {
 		return fmt.Errorf("RPC client error: %v", err)
 	}
+
+	info, err := c.GetNetworkInfo()
+
+	switch err := err.(type) {
+	case nil:
+		log.Debugf("detected btcd version: %v", info.Version)
+	// Inspect the RPC error to ensure the method was not found, otherwise
+	// we actually ran into an error.
+	case *btcjson.RPCError:
+		if err.Code != btcjson.ErrRPCMethodNotFound.Code {
+			return fmt.Errorf("unable to detect btcd version: %v", err)
+		}
+	default:
+		return fmt.Errorf("unable to detect btcd version: %v", err)
+	}
+
 	btc.c = c
 	return nil
 }
--- a/connectors/rsk.go
+++ b/connectors/rsk.go
@@ -243,7 +243,7 @@ func (rsk *RSK) GasPrice() (uint64, error) {
 		defer cancel()
 		var price *big.Int
 		price, err = rsk.c.SuggestGasPrice(ctx)
-		if price != nil && price.Cmp(big.NewInt(0)) > 0 {
+		if price != nil && price.Cmp(big.NewInt(0)) >= 0 {
 			return price.Uint64(), nil
 		}
 		time.Sleep(rpcSleep)
--- a/connectors/rsk.go
+++ b/connectors/rsk.go
@@ -36,8 +36,9 @@ const (
 )
 
 type RSKConnector interface {
-	Connect(endpoint string) error
+	Connect(endpoint string, chainId *big.Int) error
 	Close()
+	GetChainId() (*big.Int, error)
 	EstimateGas(addr string, value uint64, data []byte) (uint64, error)
 	GasPrice() (uint64, error)
 	HashQuote(q *types.Quote) (string, error)
@@ -83,7 +84,7 @@ func NewRSK(lbcAddress string, bridgeAddress string, requiredBridgeConfirmations
 	}, nil
 }
 
-func (rsk *RSK) Connect(endpoint string) error {
+func (rsk *RSK) Connect(endpoint string, chainId *big.Int) error {
 	log.Debug("connecting to RSK node on ", endpoint)
 	client, err := ethclient.Dial(endpoint)
 	if err != nil {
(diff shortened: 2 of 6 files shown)


connectors/rsk.go
```diff
@@ -243,7 +243,7 @@ func (rsk *RSK) GasPrice() (uint64, error) {
 		defer cancel()
 		var price *big.Int
 		price, err = rsk.c.SuggestGasPrice(ctx)
-		if price != nil && price.Cmp(big.NewInt(0)) > 0 {
+		if price != nil && price.Cmp(big.NewInt(0)) >= 0 {
```
aguschirico, line 246: I think 0 gas price should be only allowed when the network is RegTest, otherwise you could be starting the service under  unhealthy conditions.

connectors/bitcoin.go
```diff
@@ -84,6 +87,22 @@ func (btc *BTC) Connect(endpoint string, username string, password string) error
 	if err != nil {
 		return fmt.Errorf("RPC client error: %v", err)
 	}
+
+	info, err := c.GetNetworkInfo()
```
javicid, line 91: If I remember correctly, c.New fails if it cannot establish a connection. Why is this necessary?

connectors/bitcoin.go
```diff
@@ -84,6 +87,22 @@ func (btc *BTC) Connect(endpoint string, username string, password string) error
...
+	case *btcjson.RPCError:
+		if err.Code != btcjson.ErrRPCMethodNotFound.Code {
+			return fmt.Errorf("unable to detect btcd version: %v", err)
+		}
+	default:
+		return fmt.Errorf("unable to detect btcd version: %v", err)
+	}
+
 	btc.c = c
 	return nil
```
javicid, line 107: This can return nil even if err != nil.

connectors/rsk.go
javicid: I am not sure in which cases gas price can be 0, but this looks fine to me. Especially taking into account that this function is only used to verify the RSK connectivity.

connectors/bitcoin.go
```diff
@@ -84,6 +87,22 @@ func (btc *BTC) Connect(endpoint string, username string, password string) error
...
+
+	switch err := err.(type) {
+	case nil:
+		log.Debugf("detected btcd version: %v", info.Version)
+	// Inspect the RPC error to ensure the method was not found, otherwise
+	// we actually ran into an error.
+	case *btcjson.RPCError:
+		if err.Code != btcjson.ErrRPCMethodNotFound.Code {
+			return fmt.Errorf("unable to detect btcd version: %v", err)
+		}
```
javicid, line 101: Is it possible that the BTC node does not implement the getInfo method?

javicid (changes requested):

connectors/rsk.go
Vovchyk: 0 is being returned in `regtest`

connectors/bitcoin.go
Vovchyk: looks like it doesn't. I was able to start the server without spinning up a bitcoind node.

connectors/bitcoin.go
Vovchyk: only if `err.Code == btcjson.ErrRPCMethodNotFound.Code` (for `"method":"getnetworkinfo",`), which is a case for some older bitcoin nodes, but still if the btc node is able to return the `ErrRPCMethodNotFound` error code, it means that it's up and running

connectors/bitcoin.go
Vovchyk: new versions don't provide `getInfo`, it's deprecated. Instead there are `getnetworkinfo`, `getblockchaininfo` etc

connectors/rsk.go
aguschirico: this function is not only used to verify RSK connectivity. It's also used by the getQuoteHandler(). This means that if the node is faulty and gives 0 gas price, the quote price can be set to 0, which is an undesired effect. I'd change the function that is used for verifying the connectivity, to something like "eth_getChainId" or smth like that.

connectors/bitcoin.go
aguschirico: Indeed the fact that one was able to run the app without having a btc node available is something I was aware of. The issue ends up blowing up when the user calls the acceptQuote

connectors/rsk.go
javicid: Oh, I didn't see the use in server.go. 

The question is: can the node return 0 gas price in cases where the gas price is not actually 0? Wouldn't the node raise an error in these cases? In other words, can the node return a wrong gas price?

connectors/bitcoin.go
javicid: Ok, I didn't know this

connectors/bitcoin.go
javicid: Ok

connectors/rsk.go
aguschirico: Not sure how likely to happen in mainnet, but in testnet it happened a few times within 2 months.

Vovchyk: well, for me `0` is a valid gas price. In essence, it all depends on the node's configs, for instance, there's `minGasPrice` config for a miner. In `regtest` it's zero, that's why `eth_gasPrice` returns zero. Nothing prevents you from spinning up your own test network with this config set to zero.

javicid: @aguschirico did it return 0 because gas price was actually 0 or because there was an error computing gas price?

Vovchyk: `gasPrice` is a dynamic variable and depends on a few factors including aforementioned config, gas prices of txs of recent blocks etc., and is cached in RAM, so after restarting a node, another value may be returned

javicid: Then I think that 0 should be valid as well

javicid (dismissed):

connectors/rsk.go
aguschirico: Despite the technical reasons for this value to be 0 eventually: do we want to allow a quote to be generated with gasPrice 0 in mainnet? I don't think we want, so I'd verify connectivity to the node using a different function and/or allow this value only for any env != mainnet

javicid: But is mainnet ever going to return 0 gas price? According to this: https://stackoverflow.com/questions/55431745/how-do-geth-estimategas-and-suggest-gas-price-work

the fallback value is 1 (at least on Ethereum). Maybe it's ok to trust the value returned by the node. Otherwise, how can we tell if the gas price is valid? Could be greater than 0 and still be inaccurate.

I agree with changing the connectivity check though.

Vovchyk: > I agree with changing the connectivity check though.

I like this idea. I see `chainId` in the `config.json` file (under the `provider` section). Do you think we could trigger `eth_chainId` on an RSK node and compare the result with a value in the config and see if a proper type of the rsk node is connected?

Vovchyk: please check [this](https://github.com/rsksmart/liquidity-provider-server/pull/35/commits/d6081276c03cd6d431ecd468b0240ca382f65342) out. Used chainId endpoint for connection verification (instead of gasPrice)

aguschirico: LGTM

aguschirico (approved):

connectors/rsk.go
javicid: We could also remove chain ID from the config file if we can get it from the node.

javicid (approved):

PR #36, opened 2021-12-14
Added docker compose

Vovchyk:
Docker-compose for services involved in flyover

### DIFF
--- a/docker-compose/.env
+++ b/docker-compose/.env
@@ -1,0 +1,6 @@
+
+LIQUIDITY_PROVIDER_RSK_ADDR=0x9D93929A9099be4355fC2389FbF253982F9dF47c
+LIQUIDITY_PROVIDER_RSK_KEY={"address":"9d93929a9099be4355fc2389fbf253982f9df47c","crypto":{"cipher":"aes-128-ctr","ciphertext":"2f6b816b46ea7e9917acad870e4acb8fae1ac57535d20d9f7f5bbd6fc9ceea6f","cipherparams":{"iv":"5d57f07627bfab8e840fea6b7cbeb123"},"kdf":"scrypt","kdfparams":{"dklen":32,"n":262144,"p":1,"r":8,"salt":"5272e0072f0fb5d53171d250729979777abc65f4bd946a995397c26eef33ef92"},"mac":"f7ccbf7492edf922f11d67b5ff864d7d5b5df845586b0669af563626ba1ddad8"},"id":"f344d854-cea8-479c-93ff-0f496e9a4df3","version":3}
+LIQUIDITY_PROVIDER_RSK_KEY_PASS=test
+
+LBC_ADDR=
--- a/docker-compose/lbc-deployer/deploy.sh
+++ b/docker-compose/lbc-deployer/deploy.sh
@@ -1,0 +1,25 @@
+#!/bin/bash
+
+set -e
+
+if [[ ! -z "${LBC_ADDR}" ]]; then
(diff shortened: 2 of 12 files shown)


docker-compose/.env
```diff
@@ -0,0 +1,6 @@
+
```
josedahlquist, line 1: Should we push a .env.test file instead?

docker-compose/lbc-deployer/deploy.sh
```diff
@@ -0,0 +1,25 @@
...
+if [[ -d $CONTRACTS_DIR && "$(ls -A $CONTRACTS_DIR)" ]]; then
+    echo "Cleaning up $CONTRACTS_DIR directory..."
+    rm $CONTRACTS_DIR/*.*
+fi
+
+echo "Deploying contracts to RskJ..."
+
+cd /code/lbc
+
+npx truffle deploy --network rskRegtest
```
josedahlquist, line 21: So this will be just for regtest?

docker-compose/rskj/Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+FROM openjdk:8-jdk as runner
+
+RUN useradd -m rsk
+
+WORKDIR /home/rsk
+
+USER rsk
+COPY --from=builder --chown=rsk:rsk /code/rskj/rskj-core/build/libs/rskj-core-* ./
+
+ENTRYPOINT ["java", "-Drpc.providers.web.http.bind_address=0.0.0.0", "-Drpc.providers.web.http.hosts.0=localhost", "-Drpc.providers.web.http.hosts.1=rskj", "-cp", "rskj-core-3.1.0-IRIS-all.jar", "co.rsk.Start"]
```
josedahlquist, line 30: this is starting a mainnet node, right?

josedahlquist (commented):
I'm a bit confused here... remind me, what's the purpose of the containerization? because I see some things that makes me thing it's is for testing purpose and other that this could be intenteded for mainnet...?

Vovchyk:
> I'm a bit confused here... remind me, what's the purpose of the containerization? because I see some things that makes me thing it's is for testing purpose and other that this could be intenteded for mainnet...?

I believe the initial idea was to have a capability to easily spin up local env. Im now thinking that it'd be nice to make use of it for running a few integration tests (e.g. in the form of gihub actions), which would include a couple of use cases of quote generation and acceptance. What do you think?

docker-compose/.env
Vovchyk: do you see some other way of using it? For now, we don't have to provide `--env-file` arg for docker-compose command, as that's a default name

docker-compose/lbc-deployer/deploy.sh
Vovchyk: yeah, the idea was to spin up a local environment with regtest  instances of rsk and bitcoin

docker-compose/rskj/Dockerfile
Vovchyk: well, it's just an entry point. The network flag is being provided in the docker-compose.yml file [here](https://github.com/rsksmart/liquidity-provider-server/pull/36/files#diff-21796814c2204c04ba29b3834e57d5399c2ba04589de42b2be845cf6a0309b11R24)

josedahlquist:
I think having integration tests for this would be really great. For the time we can leave it as is and move on. Thanks Vlad!

josedahlquist (approved):

josedahlquist (approved):

PR #39, opened 2022-01-31
Added btc watcher improvements

Vovchyk:
- change how the address watcher monitors addresses to NOT stop monitoring right after the first deposit.
- address watcher should check that the deposit must have at least the negotiated amount (funds to advance + fees). Anything below that will be ignored.
- the address watcher should stop monitoring only after the time for deposit has elapsed, or the expected deposit deposit happens

### DIFF
--- a/connectors/bitcoin.go
+++ b/connectors/bitcoin.go
@@ -24,13 +24,14 @@ import (
 )
 
 type AddressWatcher interface {
-	OnNewConfirmation(txHash string, confirmations int64, amount float64)
+	OnNewConfirmation(txHash string, confirmations int64, amount btcutil.Amount)
+	OnExpire()
 	Done() <-chan struct{}
 }
 
 type BTCConnector interface {
 	Connect(endpoint string, username string, password string) error
-	AddAddressWatcher(address string, interval time.Duration, w AddressWatcher) error
+	AddAddressWatcher(address string, minAmount btcutil.Amount, interval time.Duration, exp time.Time, w AddressWatcher) error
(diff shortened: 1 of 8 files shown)


connectors/bitcoin.go
```diff
@@ -24,13 +24,14 @@ import (
 )
 
 type AddressWatcher interface {
-	OnNewConfirmation(txHash string, confirmations int64, amount float64)
+	OnNewConfirmation(txHash string, confirmations int64, amount btcutil.Amount)
+	OnExpire()
```
Vovchyk, line 28: one more callback method to notify when time for deposit has elapsed

connectors/bitcoin.go
```diff
@@ -39,8 +40,18 @@ type BTCConnector interface {
...
+	GetBlockVerbose(blockHash *chainhash.Hash) (*btcjson.GetBlockVerboseResult, error)
+	ListUnspentMinMaxAddresses(minConf, maxConf int, addrs []btcutil.Address) ([]btcjson.ListUnspentResult, error)
+	GetBlock(blockHash *chainhash.Hash) (*wire.MsgBlock, error)
+	GetRawTransaction(txHash *chainhash.Hash) (*btcutil.Tx, error)
+	Disconnect()
+}
+
 type BTC struct {
-	c       *rpcclient.Client
+	c       BTCClient
```
Vovchyk, line 54: use an interface instead of direct implementation - good for unit tests

josedahlquist (dismissed):

josedahlquist (approved):

PR #42, opened 2022-02-09
Added check for minimum pegin value on quote generation

Vovchyk:


http/server.go
```diff
@@ -276,12 +289,18 @@ func (s *Server) getQuoteHandler(w http.ResponseWriter, r *http.Request) {
 		}
 	}
 
+	if len(quotes) == 0 && amountBelowMinLockTxValue {
+		log.Error("error getting quote; requested amount below bridge's min lock tx value: ", qr.ValueToTransfer)
+		http.Error(w, "bad request; requested amount below bridge's min lock tx value", http.StatusBadRequest)
```
josedahlquist, line 294: The internal name is minLockTx as historically that was the name of the operation, but it was changed to pegin a while ago so users would be more used to read pegin rather than lock
```suggestion
		log.Error("error getting quote; requested amount below bridge's min pegin tx value: ", qr.ValueToTransfer)
		http.Error(w, "bad request; requested amount below bridge's min pegin tx value", http.StatusBadRequest)
```

josedahlquist (approved):

josedahlquist (approved):

PR #50, opened 2022-04-07
Fixed liquidity management

Vovchyk:


http/watcher.go
```diff
@@ -92,12 +94,28 @@ func (w *BTCAddressWatcher) Done() <-chan struct{} {
 func (w *BTCAddressWatcher) performCallForUser() error {
```
josedahlquist, line 94: do we need a retry logic here?

josedahlquist (approved):

PR #51, opened 2022-04-20
[devops] Add testnet CD workflow

rafaiovlabs:


.github/workflows/testnet.yml
```diff
@@ -0,0 +1,37 @@
+---
+name: liquidity provider server deploy
+
+on:
+  push:
+    branches:
+      - master
```
Vovchyk, line 7: I guess it should be `testnet` branch

Vovchyk (changes requested):
btw.. do we still need `.github/workflows)/master.yml` ?

.github/workflows/testnet.yml
rafaiovlabs: Thanks!

PR #62, opened 2022-08-22
As a user I want to be able to retrieve options for a peg-out process…

joaoiovlabs:
… from the Liquidity Provider Server And set up docker images to run in Apple M1.


http/server.go
```diff
@@ -113,6 +126,7 @@ func (s *Server) Start(port uint) error {
 	r := mux.NewRouter()
 	r.Path("/health").Methods(http.MethodGet).HandlerFunc(s.checkHealthHandler)
 	r.Path("/getQuote").Methods(http.MethodPost).HandlerFunc(s.getQuoteHandler)
+	r.Path("/getQuotePegOut").Methods(http.MethodPost).HandlerFunc(s.getQuotePegOutHandler)
```
coolwather, line 129: I think that is a best pratice if we define all pegout routes as pegout/getQuote, pegout/validateSignature.
And the name of the route should be in plural getQuotes because can be more than one quote.

http/server.go
```diff
@@ -541,3 +563,98 @@ func (s *Server) storeQuote(q *types.Quote) error {
 func getQuoteExpTime(q *types.Quote) time.Time {
 	return time.Unix(int64(q.AgreementTimestamp+q.TimeForDeposit), 0)
 }
+
```
coolwather, line 566: Same here, I think the name of the method should be in plural. Returns an array.

coolwather (approved):

PR #63, opened 2022-08-26
Accept quote

joaoiovlabs:
- Accept quote implementation (It was not taken into account security validations, all business rules should be defined);
- It was accepted LIQUIDITY PROVIDER SERVER postman collection.

### DIFF
--- a/http/server.go
+++ b/http/server.go
@@ -78,6 +78,10 @@ type acceptRes struct {
 	BitcoinDepositAddressHash string `json:"bitcoinDepositAddressHash"`
 }
 
+type acceptResPegOut struct {
+	Signature string `json:"signature"`
+}
+
 type acceptReqPegout struct {
 	QuoteHash         string `json:"quoteHash"`
 	DerivationAddress string `json:"derivationAddress"`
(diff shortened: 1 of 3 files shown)


http/server.go
```diff
@@ -73,6 +73,16 @@ type acceptReq struct {
 	QuoteHash string
 }
 
+type acceptRes struct {
+	Signature                 string `json:"signature"`
+	BitcoinDepositAddressHash string `json:"bitcoinDepositAddressHash"`
```
augustoIOV, line 78: why would we send this bitcoindepositaddresshash as part of the response? the user should have the address already in the quote. AFAIU the only necessary response value is the Signature.

http/server.go
```diff
@@ -658,3 +652,67 @@ func (s *Server) getQuotesPegOutHandler(w http.ResponseWriter, r *http.Request)
...
+
+	err = s.addAddressWatcher(quote, req.QuoteHash, req.DerivationAddress, signB, p, types.RQStateWaitingForDeposit)
+	if err != nil {
+		log.Error("error adding address watcher: ", err.Error())
+		http.Error(w, "internal server error", http.StatusInternalServerError)
+		return
+	}
+
+	signature := hex.EncodeToString(signB)
+	returnQuoteSignFunc(w, signature, req.DerivationAddress)
```
augustoIOV, line 716: seems unnecessary and risky to send the derivation address back from the request.

augustoIOV (changes requested):

coolwather (approved):

augustoIOV (approved):
Although I see integration tests (postman), remember that we discussed creating an end to end test scenario (it can be a tiny typescript/go project in a separate folder) that ensures all of the steps on the happy path are working.

PR #65, opened 2022-09-06
Fixed LGTM

coolwather:


http/server.go
```diff
@@ -223,7 +223,6 @@ func (s *Server) addAddressWatcher(quote *types.Quote, hash string, depositAddr
 		delete(s.watchers, hash)
 	})
 	if err == nil {
-		log.Info("added watcher for quote: : ", hash, "; deposit addr: ", depositAddr)
```
augustoIOV, line 226: Does it give a warning if you set la debug instead of info?

http/server.go
```diff
@@ -712,7 +711,7 @@ func (s *Server) acceptQuotePegOutHandler(w http.ResponseWriter, r *http.Request
 
 	p := getProviderByAddress(s.providers, quote.LPRSKAddr)
 
-	quoteHashInBytes, err := hex.DecodeString(req.QuoteHash)
+	quoteHashInBytes, _ := hex.DecodeString(req.QuoteHash)
```
augustoIOV, line 714: we should be checking this error, not ignoring it, right?

joaoiovlabs (approved):

PR #90, opened 2022-12-12
Added new parameter

coolwather:
Was added a new parameter to the getQuote method.
lpAddress -> The address of the liquidity provider that will do the cross chain conversion


http/server.go
```diff
@@ -331,14 +332,19 @@ func (s *Server) getQuoteHandler(w http.ResponseWriter, r *http.Request) {
 		http.Error(w, "bad request "+err.Error(), http.StatusBadRequest)
 		return
 	}
-	log.Debug("received quote request: ", fmt.Sprintf("%+v", qr))
```
MaximStanciu8, line 334: will you leave this here?

MaximStanciu8 (approved):

PR #151, opened 2023-03-02
Bitcoin BECH32 Address validation

manfioLP:
https://rsklabs.atlassian.net/browse/GBI-1148?atlOrigin=eyJpIjoiN2JjMzRlNGQ2YmZiNGVlZjlhMjNhMmQ1NWNiNWYxMzIiLCJwIjoiaiJ9

return error for bech32 type while we still dont have implementation

### DIFF
--- a/connectors/bitcoin.go
+++ b/connectors/bitcoin.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"encoding/binary"
 	"fmt"
+	"regexp"
 	"time"
 
 	"github.com/btcsuite/btcd/btcjson"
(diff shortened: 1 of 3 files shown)


connectors/bitcoin.go
```diff
@@ -817,3 +818,18 @@ func getOpCodeFromInt(val int) byte {
...
 	}
 }
+
+func isP2PKH(address string) bool {
+	pattern := regexp.MustCompile(`^[13][a-km-zA-HJ-NP-Z0-9]{25,34}$`)
+	return pattern.MatchString(address)
+}
+
+func isP2SH(address string) bool {
+	pattern := regexp.MustCompile(`^3[a-km-zA-HJ-NP-Z0-9]{25,34}$`)
```
MaximStanciu8, line 828: How about p2sh for test env? starting with 2?

MaximStanciu8 (changes requested):

connectors/bitcoin.go
manfioLP: @MaximStanciu8 case addressed

MaximStanciu8 (approved):

coolwather (approved):

PR #153, opened 2023-03-02
Feature/gbi 1105 return pegout call fee

coolwather:


http/server.go
```diff
@@ -184,12 +184,12 @@ func (s *Server) AddProvider(lp pegin.LiquidityProvider) error {
 			From:   addr,
 			Signer: lp.SignTx,
 		}
-		providerID,err := s.rsk.RegisterProvider(opts,"Provider Name",big.NewInt(10),big.NewInt(7200),big.NewInt(3600),big.NewInt(10),big.NewInt(100),"http://localhost/api",true)
+		providerID, err := s.rsk.RegisterProvider(opts, "Provider Name", big.NewInt(10), big.NewInt(7200), big.NewInt(3600), big.NewInt(10), big.NewInt(100), "http://localhost/api", true)
```
MaximStanciu8, line 187: I still dont like magic numbers in code, worth to move it in config as default values?

MaximStanciu8 (changes requested):

http/server.go
coolwather: We can create a task to that, because it's related to the provider registration, that @Dominikkq did. You can the task to me or him I believe will be a simple task, but I think it's important to keep separated.

MaximStanciu8 (approved):

PR #159, opened 2023-03-07
Feature/gbi 1140 upgrade bitcoin node

coolwather:


rafaiovlabs (commented):
Erase commented lines here, if commented and not needed we should erase them 
.

rafaiovlabs (changes requested):
Please erase commented lines on Dockerfiles. if we need a comment please add them only if there are truly necessary.

docker-compose/lps/start.sh
```diff
@@ -4,6 +4,23 @@ set -e
...
 
+echo "Testing if we have a default wallet"
+
+curl -s "http://bitcoind01:5555" --user "$BTCD_RPC_USER:$BTCD_RPC_PASS" -H "Content-Type: application/json" -d '{"jsonrpc": "1.0", "method": "getwalletinfo", "params": [], "id":"getwallet"}' | grep "{\"result\":null,\"error\":{\"code\":-18" \
+  && echo "No default wallet" \
+  && echo "Creating wallet" \
+  && curl -s "http://bitcoind01:5555" --user "$BTCD_RPC_USER:$BTCD_RPC_PASS" -H "Content-Type: application/json" -d '{"jsonrpc": "1.0", "method": "createwallet", "params": ["main", false, false, "", true, false, true], "id":"createwallet"}' \
+
+
+jq \
```
Luisfc68, line 15: I think this is not necessary because we're not using config.json anymore

it/config.json
```diff
@@ -0,0 +1,47 @@
+{
```
Luisfc68, line 1: This file is no longer needed with GBI-1123

Luisfc68 (commented):
LGTM

docker-compose/lps/start.sh
coolwather: It's better to create a task for it and do the proper test.
Just to keep track.

it/config.json
coolwather: Was added to gitignore

Luisfc68 (approved):

PR #169, opened 2023-03-10
GBI-964 + GBI-1163

Dominikkq:
Extended OpenAPI Specifications, moved default provider details to .env


connectors/bindings/lbc.go
```diff
@@ -49,16 +49,16 @@ type LiquidityBridgeContractPegOutQuote struct {
...
-	Provider                common.Address
-	Name                    string
-	Fee                     *big.Int
-	QuoteExpiration         *big.Int
-	AcceptedQuoteExpiration *big.Int
-	MinTransactionValue     *big.Int
-	MaxTransactionValue     *big.Int
-	ApiBaseUrl              string
-	Status                  bool
+	Id                      *big.Int `json:"id" example:"1" description:"Provider id"`
```
Luisfc68, line 52: I think this annotations will be lost when code is re generated

Dominikkq: The lbc.go already contains fair amount of differences from ones being generated, best practise is to copy paste functions to old lbc.go and replace the ABI to not break server from different variable names being generated.

MaximStanciu8 (approved):

PR #170, opened 2023-03-10
[GBI-1172] - Fix Healthcheck on E2E Test

manfioLP:
https://rsklabs.atlassian.net/browse/GBI-1172

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -9,6 +9,11 @@ jobs:
   e2e-test:
     runs-on: ubuntu-18.04
     steps:
+      - name: Set up Go 1.18
+        uses: actions/setup-go@v3
+        with:
+          go-version: 1.18
+
       - name: Check out repository code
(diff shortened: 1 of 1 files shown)


.github/workflows/e2e.yml
```diff
@@ -9,6 +9,11 @@ jobs:
   e2e-test:
     runs-on: ubuntu-18.04
     steps:
+      - name: Set up Go 1.18
+        uses: actions/setup-go@v3
+        with:
+          go-version: 1.18
```
Luisfc68, line 15: Maybe is better to use 1.20 version? since its the one specified in readme

manfioLP: you['re right! thanks luis :)

manfioLP: actually go-version 1.2 works a bit oddly in GH actions, I will leave it to 1.18 until future updates

rafaiovlabs (approved):
LGTM

PR #180, opened 2023-03-16
Watch for register instead of fetching receipt

Luisfc68:
Guys a couple important things about this PR:

- In order to listen to events we need to connect to RSKJ node using websocket, not with http, so I added the option to enable the websocket in the docker-compose files and updated all the urls to use websocket instad of http. Thats the reason why I included Rafael as reviewer
- Now that we are listening to the register event we could reduce the timeout that we increased yesterday during deploy right?

Task: https://rsklabs.atlassian.net/browse/GBI-1184


docker-compose/.env.regtest
```diff
@@ -5,7 +5,7 @@ IRIS_ACTIVATION_HEIGHT=10
 MAX_QUOTE_VALUE=600000000000000000
 SERVER_PORT=8080
 DB_PATH=./db/server.db
-RSK_ENDPOINT=http://rskj:4444
+RSK_ENDPOINT=ws://rskj:4445/websocket
```
coolwather, line 8: I think that we can have the 2 endpoints, sot always we will need to listen to events, and with the increase of use in the application we can overload the websocket port in server.
In my opinion right now the only point that we need the socket is in registration.
But I want to hear different ideas from team.

Luisfc68: Yes, right now the only point that we need web socket is the registration. But I don't understand why you say that we can overload web socket with the increase of use in the application, we only use one connection to the web socket in the application lifecycle and we send more or less the same amount of TCP packages that we send using HTTP (without making a TCP connection per request), the only difference is that this is a 2-way connection. The only disadvantage I really see is that if application losses connection to RSK node it need to do the handshake again but besides that I don't see other disadvantages

MaximStanciu8 (approved):

PR #185, opened 2023-03-17
Feature/GBI-1187 - Add required fields and DTOs

Luisfc68:
I only added DTOs for pegin because sdk currently doesn't have pegout. With this PR now SDK passes all the integration test with using generated code


Luisfc68:
https://rsklabs.atlassian.net/browse/GBI-1187

Dominikkq (approved):

PR #609, opened 2025-01-24
Bugfix/GBI-2451 - Await function refactor

Luisfc68:
## What
1. Update linter version
2. Modify `AwaitTxWithCtx` function to remove the defer usage and update the error messages
3. Increase mining wait timeout from 2 min to 5 min

## Why
1. To have the latest linting rules in the project. The project was updated to accomplish these rules except for `G115` of gosec which is related to the casting between int and uint. This rule is ignored for now as updating the project to satisfy it would have a very high impart
2. To prevent mistakes related to the usage of defer. The fix of this ticket was already implemented however is worth to improve this function to be more maintainable. The function was also converted to public to add the proper tests.
3. To not timeout on blocks with outlier waiting time 

P.S.
Besides updating the linter, we're making it more strict by enabling more rules. But that will come in a separate PR to reduce the noise to review this one.

## Task
https://rsklabs.atlassian.net/browse/GBI-2451


internal/adapters/dataproviders/rootstock/common.go
```diff
@@ -16,7 +16,7 @@ import (
 const (
 	rpcCallRetryMax     = 3
 	rpcCallRetrySleep   = 1 * time.Minute
-	txMiningWaitTimeout = 2 * time.Minute
+	txMiningWaitTimeout = 5 * time.Minute
```
gsoares85, line 19: Why not env var?

internal/usecases/liquidity_provider/generate_default_credentials_test.go
```diff
@@ -57,7 +57,7 @@ func TestGenerateDefaultCredentialsUseCase_Run(t *testing.T) {
 		"been generated and saved in the file %s.Please keep this file safe. The first time you open the "+
 		"management interface, you will be asked to change this password.", passwordFile)
 	assert.Contains(t, buff.String(), expectedLog)
-	assert.Positive(t, len(writtenPassword))
+	assert.NotEmpty(t, writtenPassword)
```
gsoares85, line 60: Password can be any size, even 1 letter?

MaximStanciu8 (approved):

gsoares85 (approved):

PR #632, opened 2025-02-14
Feat/GBI-2485/Add rsk-monitor as a Docker container

AndresQuijano:
## What
-Add a Docker container to use [rsk-monitor](https://github.com/rsksmart/rsk-monitor) and give metrics about BTC addresses.
-Add a command in the Makefile to easily run the container.

## Why
This is part of the delivery strategy of the monitor for Systema

[Jira tk](https://rsklabs.atlassian.net/browse/GBI-2485)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -80,8 +80,10 @@ utils-docker:
 	mkdir -p utils
 	docker build -f docker-compose/utils/Dockerfile --output=utils .
 
+MONITOR_PORT ?= 8090
 monitoring:
 	docker build -f docker-compose/monitoring/Dockerfile -t monitoring .
 	docker run \
-		-p 8080:8080 \
+		-p $(MONITOR_PORT):$(MONITOR_PORT) \
+		-e MONITOR_PORT=$(MONITOR_PORT) \
 		monitoring
--- a/Makefile
+++ b/Makefile
@@ -79,3 +79,9 @@ utils-docker:
 	rm -rf utils
 	mkdir -p utils
 	docker build -f docker-compose/utils/Dockerfile --output=utils .
+
+monitoring:
+	docker build -f docker-compose/monitoring/Dockerfile -t monitoring .
+	docker run \
+		-p 8080:8080 \
+		monitoring
--- a/docker-compose/monitoring/src/config.ts
+++ b/docker-compose/monitoring/src/config.ts
@@ -1,13 +1,9 @@
 import { type BitcoinAddress } from '@rsksmart/rsk-monitor'
 
-
-/**
- * Uncomment this to monitor testnet addresses
- */
 export const MONITORED_ADDRESSES: BitcoinAddress[] = [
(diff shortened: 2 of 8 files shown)


docker-compose/monitoring/src/config.ts
```diff
@@ -0,0 +1,32 @@
...
+]
+
+export const MONITOR_CONFIG = {
+  pollingIntervalSeconds: 10,
+  monitorName: 'bitcoin-balance-monitor',
+  network: 'testnet' as 'mainnet' | 'testnet'
+} 
+
+
+/**
```
MaximStanciu8, line 20: Remove this comment maybe?

AndresQuijano: I wanted to leave it there so it is easier to configure. This comment is even in the documentation as an option/example of how to configure mainnet addresses

MaximStanciu8 (approved):

docker-compose/monitoring/src/config.ts
AndresQuijano: Changed.

Makefile
```diff
@@ -79,3 +79,9 @@ utils-docker:
 	rm -rf utils
 	mkdir -p utils
 	docker build -f docker-compose/utils/Dockerfile --output=utils .
+
+monitoring:
+	docker build -f docker-compose/monitoring/Dockerfile -t monitoring .
+	docker run \
+		-p 8080:8080 \
```
Luisfc68, line 86: this port is the same as for the LPS, since we're delivering in the same compose there is a change that they use this container in the same vm as the LPS so I'd suggest to use a different port

docker-compose/monitoring/src/index.ts
```diff
@@ -0,0 +1,24 @@
...
+import { MONITORED_ADDRESSES, MONITOR_CONFIG } from './config'
+
+async function main (): Promise<void> {
+  await BitcoinMempoolMonitorBuilder.create({
+    pollingIntervalSeconds: MONITOR_CONFIG.pollingIntervalSeconds,
+    monitorName: MONITOR_CONFIG.monitorName,
+    network: MONITOR_CONFIG.network
+  } as MonitorConfig, MONITORED_ADDRESSES)
+    .withBalanceMetric()
+    .withExporters(new ConsoleExporter(), new PrometheusExporter(8080, 'bitcoinbalancemonitor'))
```
Luisfc68, line 11: same comment about port

Luisfc68 (changes requested):

Makefile
AndresQuijano: Thanks. Changed.

docker-compose/monitoring/src/index.ts
AndresQuijano: Thanks. Changed.

Luisfc68 (approved):

PR #651, opened 2025-03-05
Percentage-based fees | GBI-2527 | GBI-2531 | GBI-1838

Dominikkq:
**What:**
Add input fields for percentage-based fees(fixedFee, percentatageFee) in the UI
Ensured proper field validation for new values
Added optional enable/disable buttons on fixedFee and percentatageFee
Added warning message when both fixedFee and percentatageFee are disabled
Split up and refactored createInput for clarity
Added validation for percentage fee and ZeroOrPositiveString values
Updated lps-env-sh.sh script to log failed request and error to console for debugging
Removed callFee and added feePercentage and fixedFee in lps-env-sh.sh config setup
Updated encoding.go to support proper percentageFee parsing

**Tasks:**
[GBI-2527](https://rsklabs.atlassian.net/browse/GBI-2527)
[GBI-1838](https://rsklabs.atlassian.net/browse/GBI-1838)
[GBI-2531](https://rsklabs.atlassian.net/browse/GBI-2531)


internal/adapters/entrypoints/rest/common.go
```diff
@@ -28,13 +30,57 @@ func PositiveStringValidationRule(value string) bool {
 	return bigIntValue.Cmp(big.NewInt(0)) > 0
 }
 
+func ZeroOrPositiveStringValidationRule(value string) bool {
+	bigIntValue := new(big.Int)
+	bigIntValue.SetString(value, 10)
```
gsoares85, line 35: We should do error handling on the SetString. To avoid runtime errors.

gsoares85: Same on the ones bellow

pkg/liquidity_provider.go
```diff
@@ -148,6 +149,35 @@ func FromPegoutConfigurationDTO(dto PegoutConfigurationDTO) liquidity_provider.P
 	}
 }
 
+func ToPeginConfigurationDTO(config liquidity_provider.PeginConfiguration) PeginConfigurationDTO {
+	feePercentage, _ := config.FeePercentage.Native().Float64()
```
gsoares85, line 153: Should do error handling here, avoid ignore errors.

pkg/liquidity_provider.go
```diff
@@ -148,6 +149,35 @@ func FromPegoutConfigurationDTO(dto PegoutConfigurationDTO) liquidity_provider.P
...
+		PenaltyFee:     config.PenaltyFee.AsBigInt().String(),
+		FixedFee:       config.FixedFee.AsBigInt().String(),
+		FeePercentage:  feePercentage,
+		MaxValue:       config.MaxValue.AsBigInt().String(),
+		MinValue:       config.MinValue.AsBigInt().String(),
+	}
+}
+
+func ToPegoutConfigurationDTO(config liquidity_provider.PegoutConfiguration) PegoutConfigurationDTO {
+	feePercentage, _ := config.FeePercentage.Native().Float64()
```
gsoares85, line 167: Same

internal/adapters/entrypoints/rest/assets/static/management.js
```diff
@@ -303,25 +424,40 @@ function getRegularConfig(sectionId) {
...
+            if (isfeePercentageKey(key)) {
+                value = 0;
+            } else {
+                value = '0';
             }
         } else {
-            value = input.value;
-            if (!isNaN(value) && !isNaN(parseFloat(value))) {
-                value = Number(value);
+            if (isFeeKey(key)) {
```
Luisfc68, line 437: I left a comment about this function in #648 saying that it is too much nesting here, we need to make this function more maintainable, this would pop up if we had a linter for the template. I think the options here are either to refactor the function or to move the content of the else statement to a different one

internal/entities/utils/encoding.go
```diff
@@ -24,6 +26,26 @@ func NewBigFloat(value *big.Float) *BigFloat {
 	return (*BigFloat)(value)
 }
 
+func (bf *BigFloat) MarshalJSON() ([]byte, error) {
+	if bf == nil {
+		return []byte("null"), nil
+	}
+	valueStr := bf.Native().Text('f', -1)
+	return json.Marshal(valueStr)
```
Luisfc68, line 34: I think this will marshall as string, I think is better to marshall this field as a number as the range is quite low [0;100]. Also this is what is forcing you to implement a custom validator

internal/adapters/entrypoints/rest/common.go
```diff
@@ -28,13 +30,57 @@ func PositiveStringValidationRule(value string) bool {
...
+	return bigIntValue.Cmp(big.NewInt(0)) >= 0
+}
+
+func NonNegativeStringValidationRule(value string) bool {
+	bigIntValue := new(big.Int)
+	bigIntValue.SetString(value, 10)
+	return bigIntValue.Cmp(big.NewInt(0)) >= 0
+}
+
+func percentageFeeValidator(fl validator.FieldLevel) bool {
```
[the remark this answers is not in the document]
Dominikkq: Don't think I can do decimal validation inline with the library

MaximStanciu8 (approved):

internal/adapters/entrypoints/rest/assets/static/management.js
Dominikkq: Split up createInput function for maintainability after your comment on last PR, but current getRegularConfig implementation is only 43 lines long.

gsoares85 (approved):

pkg/liquidity_provider.go
```diff
@@ -59,8 +60,8 @@ type PegoutConfigurationDTO struct {
 	TimeForDeposit       uint32  `json:"timeForDeposit" validate:"required"`
 	ExpireTime           uint32  `json:"expireTime" validate:"required"`
 	PenaltyFee           string  `json:"penaltyFee" validate:"required,numeric,positive_string"`
-	FixedFee             string  `json:"fixedFee" validate:"required,numeric,positive_string"`
-	FeePercentage        float64 `json:"feePercentage" validate:"required,numeric,gt=0"`
+	FixedFee             string  `json:"fixedFee" validate:"required,numeric,min=0"`
```
Luisfc68, line 63:
```suggestion
	FixedFee             string  `json:"fixedFee" validate:"required,numeric,positive_string"`
```

Luisfc68: in this case you can use `positive_string ` that validator is the same as min=0 but for string fields, the reason why this field and the others that are amounts are strings is because the value in them might be bigger than a `uint64`

internal/adapters/dataproviders/liquidity_provider_test.go
```diff
@@ -537,6 +539,68 @@ func TestLocalLiquidityProvider_PegoutConfiguration(t *testing.T) {
 	})
 }
 
+func TestLocalLiquidityProvider_ProviderDTOValidation(t *testing.T) {
```
Luisfc68, line 542: this test is in the wrong package, should be in `pkg`

internal/adapters/dataproviders/liquidity_provider_test.go
```diff
@@ -537,6 +539,68 @@ func TestLocalLiquidityProvider_PegoutConfiguration(t *testing.T) {
...
+		expectedDTO := pkg.PeginConfigurationDTO{
+			TimeForDeposit: config.TimeForDeposit,
+			CallTime:       config.CallTime,
+			PenaltyFee:     config.PenaltyFee.AsBigInt().String(),
+			FixedFee:       config.FixedFee.AsBigInt().String(),
+			FeePercentage:  feePercentage,
+			MaxValue:       config.MaxValue.AsBigInt().String(),
+			MinValue:       config.MinValue.AsBigInt().String(),
+		}
+		assert.Equal(t, expectedDTO, dto)
```
Luisfc68, line 600: I'd also add `test.AssertNonZeroValues(t, dto)`

pkg/liquidity_provider.go
```diff
@@ -45,8 +46,8 @@ type PeginConfigurationDTO struct {
 	TimeForDeposit uint32  `json:"timeForDeposit" validate:"required"`
 	CallTime       uint32  `json:"callTime" validate:"required"`
 	PenaltyFee     string  `json:"penaltyFee" validate:"required,numeric,positive_string"`
-	FixedFee       string  `json:"fixedFee" validate:"required,numeric,positive_string"`
-	FeePercentage  float64 `json:"feePercentage" validate:"required,numeric,gt=0"`
+	FixedFee       string  `json:"fixedFee" validate:"required,numeric,min=0"`
```
Luisfc68, line 49:
```suggestion
	FixedFee       string  `json:"fixedFee" validate:"required,numeric,positive_string"`
```

Luisfc68: same here, when is a wei field you should use `positive_string` instead of `min=0`

internal/adapters/entrypoints/rest/common.go
```diff
@@ -24,17 +27,43 @@ var RequestValidator = validator.New(validator.WithRequiredStructEnabled())
...
 	bigIntValue := new(big.Int)
-	bigIntValue.SetString(value, 10)
+	_, ok := bigIntValue.SetString(value, 10)
+	if !ok {
+		return false
+	}
 	return bigIntValue.Cmp(big.NewInt(0)) > 0
 }
 
+func decimalPlacesValidator(fl validator.FieldLevel) bool {
```
Luisfc68, line 37: I'd recommend to test this validator too, because the implementation is not straight forward and we want to make sure that 1e-9 is enough in case the floating point operations result in more decimals than expected

internal/adapters/dataproviders/liquidity_provider_test.go
Luisfc68: (for both pegin and pegout)

pkg/liquidity_provider.go
Dominikkq: 0 won't be included then(fails validations)

Luisfc68 (approved):

PR #652, opened 2025-03-05
feat: added pegin report endpoint

gsoares85:
## What

Added a new endpoint under management API to return a financial report over the pegins.
- Number of Quotes Finalized -> Only the ones with the status RegisterPegInSucceeded
- Minimum and maximum Value for the quotes
- Average Value
- Total Call Fess
- Average Call Fess

Already added the average call fees since we have plans to move to a percentage fee


MaximStanciu8 (approved):

internal/usecases/pegin/get_pegin_report.go
Luisfc68, line 1: two general comments about this file:
1. just fyi, there are a couple things done here that are already included in the standard library https://pkg.go.dev/slices
2. the implementation of this use case isn't very optimal in the sense that you're iterating the same dataset 6 times and one of those times is to fetch each quote from the db (that one is the most concerning IMO), since this endpoint is expected to be expensive, because it fetches all the successful quotes from the db, I would advise to refactor this use case and to implement something to bulk read the quotes from the db, we're talking about around 110 database reads per endpoint execution in mainnet, and in testnet is way more. So implementing it in this way is going to lead probably to timeout or really long execution times

internal/usecases/pegin/get_pegin_report_test.go
```diff
@@ -0,0 +1,96 @@
...
+
+func TestGetPeginReportUseCase_Run(t *testing.T) {
+	ctx := context.Background()
+
+	retainedQuotes := []quote.RetainedPeginQuote{
+		{QuoteHash: "hash1"},
+		{QuoteHash: "hash2"},
+	}
+
+	peginQuotes := []quote.PeginQuote{
```
Luisfc68, line 22: please extend this dataset, having only 2 elements for a functionality that calculates min, max, avg, etc. Is too few

Luisfc68 (changes requested):

internal/usecases/pegin/get_pegin_report_test.go
gsoares85: done

internal/usecases/pegin/get_pegin_report.go
gsoares85: done

internal/usecases/pegin/get_pegin_report.go
```diff
@@ -0,0 +1,162 @@
...
+		MinimumQuoteValue:  minimumQuoteValue,
+		MaximumQuoteValue:  maximumQuoteValue,
+		AverageQuoteValue:  averageQuoteValue,
+		TotalFeesCollected: totalFeesCollected,
+		AverageFeePerQuote: averageFeePerQuote,
+	}, nil
+}
+
+func (useCase *GetPeginReportUseCase) calculateMinimumQuoteValue(quotes []quote.PeginQuote) *entities.Wei {
+	minimum := quotes[0].Value
```
Dominikkq, line 89: accessing quotes[0] without first checking if the quotes slice is empty will probably cause an index out of range error

internal/usecases/pegin/get_pegin_report.go
```diff
@@ -0,0 +1,162 @@
...
+				minimum = q.Value
+			}
+		}
+	}
+
+	return minimum
+}
+
+func (useCase *GetPeginReportUseCase) calculateMaximumQuoteValue(quotes []quote.PeginQuote) *entities.Wei {
+	maximum := quotes[0].Value
```
Dominikkq, line 103: same possible index out of range error here

internal/adapters/dataproviders/database/mongo/pegin.go
```diff
@@ -68,6 +68,39 @@ func (repo *peginMongoRepository) GetQuote(ctx context.Context, hash string) (*q
...
+		}
+	}
+
+	collection := repo.conn.Collection(PeginQuoteCollection)
+	filter := bson.M{"hash": bson.M{"$in": hashes}}
+
+	quotesReturn := make([]quote.PeginQuote, 0)
+
+	cursor, err := collection.Find(dbCtx, filter)
+	if errors.Is(err, mongo.ErrNoDocuments) {
```
Dominikkq, line 88: Find() never returns mongo.ErrNoDocuments

Dominikkq (changes requested):

internal/usecases/pegin/get_pegin_report.go
gsoares85: Done

internal/usecases/pegin/get_pegin_report.go
gsoares85: Done

internal/adapters/dataproviders/database/mongo/pegin.go
Luisfc68: +1

internal/adapters/dataproviders/database/mongo/pegin.go
```diff
@@ -68,6 +68,39 @@ func (repo *peginMongoRepository) GetQuote(ctx context.Context, hash string) (*q
...
+		return nil, nil
```
Luisfc68, line 89: I'd return the empty array in here, if you don't include an error, the consumer will asume the result has something in it

internal/adapters/dataproviders/database/mongo/pegin_test.go
```diff
@@ -338,3 +338,45 @@ func TestPeginMongoRepository_DeleteQuotes(t *testing.T) {
...
+		invalidHashes := []string{"invalidHash"}
+		conn := mongo.NewConnection(client, time.Duration(1))
+		repo := mongo.NewPeginMongoRepository(conn)
+
+		_, err := repo.GetQuotes(context.Background(), invalidHashes)
+		require.Error(t, err)
+		assert.Equal(t, "invalid quote hash length: expected 64 characters, got 11", err.Error())
+	})
+
+	t.Run("No quotes found", func(t *testing.T) {
```
Luisfc68, line 369: I'd change this test from no quotes found to error reading from db or similar

internal/usecases/pegin/get_pegin_report.go
Luisfc68: still see the access without the check

Luisfc68 (changes requested):

Luisfc68:
there are some unresolved comments from the previous review, if you add the check to the array length in the use case then the next if is not needed because you know the length will be at least 1

internal/usecases/pegin/get_pegin_report.go
gsoares85: Will be on my next commit, just marked here to not forgot

internal/usecases/pegin/get_pegin_report.go
```diff
@@ -0,0 +1,170 @@
...
+	}, nil
+}
+
+func (useCase *GetPeginReportUseCase) calculateMinimumQuoteValue(quotes []quote.PeginQuote) *entities.Wei {
+	if len(quotes) == 0 {
+		return entities.NewWei(0)
+	}
+	minimum := quotes[0].Value
+
+	if len(quotes) > 1 {
```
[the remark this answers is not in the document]
gsoares85: Done

internal/usecases/pegin/get_pegin_report.go
```diff
@@ -0,0 +1,170 @@
...
+	return minimum
+}
+
+func (useCase *GetPeginReportUseCase) calculateMaximumQuoteValue(quotes []quote.PeginQuote) *entities.Wei {
+	if len(quotes) == 0 {
+		return entities.NewWei(0)
+	}
+	maximum := quotes[0].Value
+
+	if len(quotes) > 1 {
```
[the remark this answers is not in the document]
gsoares85: Done

Dominikkq (approved):

PR #654, opened 2025-03-06
feat: add call fee calculation

Luisfc68:
## What
- Added call fee calculation according to the product requirement
- Log each part of the calculation with debug level

## Why
To use the actual `feePercentage` and `fixedFee` defined by the LP in the quote generation

P.S.: feel free to propose numbers for other tests cases for the calculation, I tried to cover different rounding scenarios

## Task
https://rsklabs.atlassian.net/browse/GBI-2528


MaximStanciu8 (approved):

internal/entities/quote/common.go
```diff
@@ -44,6 +46,38 @@ func ValidateQuoteHash(hash string) error {
...
+	result.Add(percentageFee, config.GetFixedFee())
+
+	log.Debugf("Percentage fee: %v%% of %v = %v", config.GetFeePercentage(), amount, percentageFee)
+	log.Debugf("Fixed fee: %v", config.GetFixedFee())
+	log.Debugf("Call fee: %v + %v = %v", percentageFee, config.GetFixedFee(), result)
+	return result
+}
+
+func calculatePercentageFee(amount *entities.Wei, percentage *utils.BigFloat) *entities.Wei {
+	const scale = 1000 // the scale needs to have at least as many zeros as the amount of decimals we want to support in the percentage
```
gsoares85, line 61: If this is configurable and the LPs can support different number of decimals, this one should be an env var.

internal/entities/quote/common.go
```diff
@@ -44,6 +46,38 @@ func ValidateQuoteHash(hash string) error {
 }
 
 func CalculateCallFee(amount *entities.Wei, config PegConfiguration) *entities.Wei {
-	// TODO implement in GBI-2528
-	return entities.NewWei(100000000000)
+	result := new(entities.Wei)
+
+	percentageFee := calculatePercentageFee(amount, config.GetFeePercentage())
```
gsoares85, line 51: If the config GetFeePercentage() return nil or invalid the method can panic.

internal/entities/quote/common.go
```diff
@@ -44,6 +46,38 @@ func ValidateQuoteHash(hash string) error {
...
+func calculatePercentageFee(amount *entities.Wei, percentage *utils.BigFloat) *entities.Wei {
+	const scale = 1000 // the scale needs to have at least as many zeros as the amount of decimals we want to support in the percentage
+	amountAsRat := new(big.Rat).SetInt(amount.AsBigInt())
+	floatPercentage, _ := percentage.Native().Float64()
+
+	percentageAsFraction := new(big.Rat).SetFrac(
+		big.NewInt(int64(floatPercentage*scale)), // Scale to avoid precision loss
+		big.NewInt(100*scale),
+	)
+	percentageFee := new(big.Rat).Mul(amountAsRat, percentageAsFraction)
```
gsoares85, line 69: There is any check for integer overflow on this multiplication?

internal/entities/quote/common.go
Luisfc68: is not configurable, is 2 decimals as per product requirement

internal/entities/quote/common.go
Luisfc68: thanks, just added those cases to the test suite

internal/entities/quote/common.go
Luisfc68: no, the big package allocates memory dynamically instead of using a fixed memory size like the native types https://pkg.go.dev/math/big#NewRat

gsoares85 (approved):

PR #661, opened 2025-03-17
feat: added pegout report endpoint

gsoares85:
###What
Added a new endpoint under management API to return a financial report over the pegouts.

Number of Quotes Finalized -> Only the ones with the status RegisterPegOutSucceeded
Minimum and maximum Value for the quotes
Average Value
Total Call Fess
Average Call Fess
Already added the average call fees since we have plans to move to a percentage fee

Task
https://rsklabs.atlassian.net/browse/GBI-2539

### DIFF
--- a/OpenApi.yml
+++ b/OpenApi.yml
@@ -990,6 +990,13 @@ paths:
         "200":
           description: ""
       summary: Get Pegin Reports
+  /reports/pegout:
+    get:
+      description: ' Get the last pegouts on the API. Included in the management API.'
+      responses:
+        "200":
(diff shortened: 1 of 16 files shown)


MaximStanciu8 (approved):

Luisfc68 (approved):
The code looks fine, but I'd like to remind a couple things:
- Now we're adding tests to the http handlers as well so they should be included in the PRs too
- These reports endpoint don't consider the gas fee nor the actual gas cost of the different transactions. This will make their results inaccurate, specially in the long term

PR #678, opened 2025-04-16
feat: DevX Automation Documentation Preparation

Dominikkq:
**What:**
Formating docs to work with docosaurus
Unifying existing documentation from devportal with github docs

**Task:**
[GBI-2572](https://rsklabs.atlassian.net/browse/GBI-2572)


MaximStanciu8 (approved):

docker-compose/README.md
```diff
@@ -1,19 +1,31 @@
-# docker-compose for Liquidity Provider Server
+# Docker Setup for Liquidity Provider Server
 
-This docker-compose files can be used to quickly spin up an environment with the server and its
-dependant services (`bitcoind` and `rskj`) for either `regtest` or `testnet`
+The provided docker-compose files can be used to quickly spin up an environment with the Liquidity Provider Server and its dependent services (`bitcoind` and `rskj`) for either `regtest` or `testnet`.
```
Luisfc68, line 3: I would clarify in this section that it is not the same file for both, we had this confusion with Systema, because the regtest env requires a regtest federation and a localstack so the setup is a bit different, for example, only the local env has the script to spin up the environment, testnet or mainnet should be done directly with the compose files

docs/DESIGN.md
```diff
@@ -10,29 +10,29 @@ liquidity providers (LPs) that store their BTC in RSK and Bitcoin. The first ver
...
 
 
-# Workflow:
+## Workflow
 
 The following diagram shows the interaction of the different components during the peg-in process. Note that the call
 to isOperational could be performed at a different time:
 
-![Flyover Sequence Diagram](diagrams/flyover-sd.png)
+![Flyover Sequence Diagram](https://raw.githubusercontent.com/rsksmart/liquidity-provider-server/refs/heads/flyover-2.3.0/docs/diagrams/flyover-sd.png)
```
Luisfc68, line 18: Question: does the dev portal support mermaid? because we could point to that file instead of the image

Luisfc68: also this is pointing to the 2.3.0 branch

docs/DESIGN.md
```diff
@@ -10,29 +10,29 @@ liquidity providers (LPs) that store their BTC in RSK and Bitcoin. The first ver
...
 
 
 The following diagrams show the interactions between Liquidity Provider, Liquidity Bridge Contract and RSK Bridge Contract in three different scenarios: 
 1. Basic fast bridge workflow (happy path)
 2. Unsuccessful call on behalf of a user
 3. Liquidity Provider fails to deliver funds to LBC
 
-![Flyover Sequence Diagram](diagrams/flyover-ad-basic.png)
+![Flyover Sequence Diagram](https://raw.githubusercontent.com/rsksmart/liquidity-provider-server/flyover-2.3.0/docs/diagrams/flyover-ad-basic.png)
```
Luisfc68, line 26: this one and the ones below are pointing to the 2.3.0 too

docs/Environment.md
Luisfc68, line 1: question: why can't we keep the table format?

docs/Environment.md
```diff
@@ -1,66 +1,69 @@
...
+| `PEGOUT_DEPOSIT_CHECK_TIMEOUT` | The time in seconds that the LPS will wait for every pegout deposit check repetition. If not provided default value will be the one defined in timeout.go. | `10` | NO |
 
 ## AWS variables
-You may notice that in [sample-config.env](../sample-config.env) there are some environment variables that are related to AWS. These variables are required to use AWS services, however, they are not listed in the table as the AWS SDK has the functionality to load them from multiple sources. For that reason, they are not accessed directly from the code and are not listed in the table above.
+You may notice that in `sample-config.env` there are some environment variables that are related to AWS. These variables are required to use AWS services, however, they are not listed in the table as the AWS SDK has the functionality to load them from multiple sources. For that reason, they are not accessed directly from the code and are not listed in the table above.
```
Luisfc68, line 66: can we maintain the link?

docs/LP-Management.md
Luisfc68, line 1: some links here are pointing to 2.3.0 branch too

Luisfc68 (commented):
nice work

README.md
```diff
@@ -16,52 +25,146 @@ To run the project locally you can follow these steps:
...
+If you want to play with the code and make modifications to it then run the following commands (remember that you need to have Go installed with the version specified in the `go.mod` file):
 1. `git clone git@github.com:rsksmart/liquidity-provider-server.git`
 2. `make tools`
 
-## Configuration
+## LPS APIs
+
+The LPS has two main APIs:
+* User/Public API: This API is used by the user to interact with the LP to agree on a quote, the API can be accessed via the Flyover dApp. See [Using the Public API](#using-the-public-api)
+* LP/Management API: This API is used by the LP to interact with the LPS to perform management operations. It can be accessed via `<LPS URL>/management`. See [LP Management](./management.md) Section
```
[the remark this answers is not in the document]
Dominikkq: It's called management.md on devportal but the point is that this specific part can be made accessible to both GitHub and DevPortal by either using a regex replacement in the pipeline or, on a second iteration of the docs linking it to the already deployed docs url. For github the file is in the /docs/ folder, which doesn't exist on devdocs which is causing this conflict. I'm very open to suggestions here.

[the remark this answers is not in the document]
Luisfc68: regex makes sense

Luisfc68 (approved):

PR #690, opened 2025-04-29
feat: Trusted Account Model + UI | GBI-2612 + GBI-2616 + GBI-2617

Dominikkq:
**What:**
Added Trusted Account data models and schemas
Implemented management endpoints for trusted accounts
Added MongoDB repository for trusted accounts
Management UI integration

**Tasks:**
[GBI-2612](https://rsklabs.atlassian.net/browse/GBI-2612)
[GBI-2616](https://rsklabs.atlassian.net/browse/GBI-2616)
[GBI-2617](https://rsklabs.atlassian.net/browse/GBI-2617)


OpenApi.yml
```diff
@@ -735,6 +757,51 @@ paths:
...
+          description: ""
+      summary: Add Trusted Account
+  /management/trusted-accounts/delete:
+    post:
+      description: ' Deletes a trusted account'
+      requestBody:
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/'
```
MaximStanciu8, line 799: missed schema

internal/adapters/entrypoints/rest/registry/registry.go
```diff
@@ -35,4 +35,6 @@ type UseCaseRegistry interface {
 	GetPegoutStatusUseCase() *pegout.StatusUseCase
 	GetAvailableLiquidityUseCase() *liquidity_provider.GetAvailableLiquidityUseCase
 	GetServerInfoUseCase() *liquidity_provider.ServerInfoUseCase
+	GetTrustedAccountsUseCase() *liquidity_provider.GetTrustedAccountsUseCase
+	SetTrustedAccountUseCase() *liquidity_provider.SetTrustedAccountUseCase
```
MaximStanciu8, line 39: Cant find  Unit Test for it

internal/adapters/entrypoints/rest/assets/management.html
```diff
@@ -105,6 +135,43 @@ <h5 class="card-title">Current Configuration</h5>
...
+                    </div>
+                    <div class="mb-3">
+                        <label for="rbtc_locking_cap" class="form-label">rBTC Locking Cap</label>
+                        <input type="text" class="form-control" id="rbtc_locking_cap" placeholder="Enter rBTC locking cap">
+                        <div class="form-text">Maximum amount of rBTC that can be locked by this account</div>
+                    </div>
+                </div>
+                <div class="modal-footer">
+                    <button type="button" class="btn btn-secondary" data-bs-dismiss="modal">Cancel</button>
+                    <button type="button" class="btn btn-primary" id="saveAccountButton">Save</button>
```
MaximStanciu8, line 169: Can we prevent Duplicate Adds Without Reload
Currently, the UI re-fetches accounts after each change but does not prevent accidental double-clicks.

internal/adapters/dataproviders/database/mongo/liquidity_provider.go
```diff
@@ -12,6 +13,7 @@ import (
 )
 
 const LiquidityProviderCollection = "liquidityProvider"
+const TrustedAccountCollection = "trustedAccounts"
```
Luisfc68, line 16: if its a diff collection then you should create a new file for the repository

internal/adapters/dataproviders/database/mongo/liquidity_provider.go
```diff
@@ -99,6 +105,92 @@ func (repo *lpMongoRepository) UpsertCredentials(ctx context.Context, credential
...
+	logDbInteraction(Read, accounts)
+	return accounts, nil
+}
+
+func (repo *lpMongoRepository) AddTrustedAccount(ctx context.Context, account liquidity_provider.TrustedAccountDetails) error {
+	dbCtx, cancel := context.WithTimeout(ctx, repo.conn.timeout)
+	defer cancel()
+	collection := repo.conn.Collection(TrustedAccountCollection)
+	existingAccount, err := repo.GetTrustedAccount(ctx, account.Address)
+	if err == nil && existingAccount != nil {
```
Luisfc68, line 146: Besides this, you should also create a unique index, like the one we have for the pegout deposits

internal/adapters/dataproviders/database/mongo/liquidity_provider.go
```diff
@@ -99,6 +105,92 @@ func (repo *lpMongoRepository) UpsertCredentials(ctx context.Context, credential
...
+		return err
+	}
+	filter := bson.M{"address": account.Address}
+	opts := options.Update().SetUpsert(true)
+	update := bson.M{"$set": account}
+	_, err = collection.UpdateOne(dbCtx, filter, update, opts)
+	if err != nil {
+		return err
+	}
+	logDbInteraction(Insert, account)
```
Luisfc68, line 174: this is an update

internal/entities/liquidity_provider/configuration.go
```diff
@@ -47,7 +48,7 @@ type HashedCredentials struct {
 }
 
 type ConfigurationType interface {
-	PeginConfiguration | PegoutConfiguration | GeneralConfiguration | HashedCredentials
+	PeginConfiguration | PegoutConfiguration | GeneralConfiguration | HashedCredentials | TrustedAccountDetails
```
Luisfc68, line 51: This is ok, but if you decide to count a trusted account as configuration then you don't need to implement a separate signature function

internal/entities/liquidity_provider/trusted_account.go
```diff
@@ -0,0 +1,30 @@
...
+import (
+	"context"
+	"errors"
+
+	"github.com/rsksmart/liquidity-provider-server/internal/entities"
+)
+
+var (
+	ErrTrustedAccountNotFound = errors.New("trusted account not found")
+	ErrDuplicateAddress       = errors.New("address already exists")
```
Luisfc68, line 12: I make this suggestion because we're in the trusted account context here, if you want this error to be generic that is also fine but then should be in the common file
```suggestion
	ErrDuplicateTrustedAccount       = errors.New("trusted account already exists")
```

internal/entities/liquidity_provider/trusted_account.go
```diff
@@ -0,0 +1,30 @@
...
+)
+
+type TrustedAccountDetails struct {
+	Address          string        `json:"address" bson:"address" validate:"required"`
+	Name             string        `json:"name" bson:"name" validate:"required"`
+	Btc_locking_cap  *entities.Wei `json:"btc_locking_cap" bson:"btc_locking_cap" validate:"required"`
+	Rbtc_locking_cap *entities.Wei `json:"rbtc_locking_cap" bson:"rbtc_locking_cap" validate:"required"`
+	Signature        string        `json:"signature" bson:"signature"`
```
Luisfc68, line 20: Signature and hash are not needed to be included in the signed struct, because they are included in the `entities.Signed` generic struct, you can use as example `PeginConfiguration`

internal/adapters/entrypoints/rest/assets/static/configUtils.js
```diff
@@ -106,6 +107,15 @@ function hasDuplicateConfirmationAmounts(confirmationArray) {
     return uniqueAmounts.size < amounts.length;
 }
 
+const formatCap = (value, unit) => {
+    try {
+        const num = parseFloat(value);
+        return parseFloat(num.toFixed(4)).toString() + ' ' + unit;
+    } catch (e) {
+        return value + ' ' + unit;
```
Luisfc68, line 115: shouldn't you show the error in this case? how can you be sure that value and unit are correct values if the try block failed?

Luisfc68 (changes requested):
none of the use cases nor the handlers are tested, please remember the unit test are mandatory

internal/adapters/entrypoints/rest/assets/management.html
Dominikkq: We are not reloading the website, we are refetch to validate that the addition was successful and to make sure the table is properly synced, and no need to protect from double clicks since all actions are protected, for ex. you can't double add an address because it will show error for trying to add duplicate address, and for deletion there is a also a an alert that requires user to confirm action, and you can't double delete the same address.

pkg/liquidity_provider.go
Luisfc68, line 1: not tested

internal/adapters/entrypoints/rest/handlers/add_trusted_account.go
```diff
@@ -0,0 +1,47 @@
...
+			return
+		}
+		accountDetails := lp.TrustedAccountDetails{
+			Address:          request.Address,
+			Name:             request.Name,
+			Btc_locking_cap:  entities.NewBigWei(request.BtcLockingCap),
+			Rbtc_locking_cap: entities.NewBigWei(request.RbtcLockingCap),
+		}
+		err = useCase.Run(req.Context(), accountDetails)
+		if err != nil {
```
Luisfc68, line 40: What about duplicated addressed? That is a known error and it would be returning 500 in this case

internal/adapters/entrypoints/rest/handlers/delete_trusted_account.go
```diff
@@ -0,0 +1,34 @@
...
+// NewDeleteTrustedAccountHandler
+// @Title Delete Trusted Account
+// @Description Deletes a trusted account
+// @Param TrustedAccountAddressRequest body pkg.TrustedAccountAddressRequest true "Address of the trusted account to delete"
+// @Success 204 object
+// @Route /management/trusted-accounts [delete]
+func NewDeleteTrustedAccountHandler(useCase *lpuc.DeleteTrustedAccountUseCase) http.HandlerFunc {
+	return func(w http.ResponseWriter, req *http.Request) {
+		var err error
+		request := &pkg.TrustedAccountAddressRequest{}
```
Luisfc68, line 20: Its better to avoid using the body in a DELETE request, https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Methods/DELETE prefer using path variables in this case

internal/adapters/entrypoints/rest/handlers/delete_trusted_account.go
```diff
@@ -0,0 +1,34 @@
...
+		if err = rest.DecodeRequest(w, req, request); err != nil {
+			return
+		} else if err = rest.ValidateRequest(w, request); err != nil {
+			return
+		}
+		err = useCase.Run(req.Context(), request.Address)
+		if err != nil {
```
Luisfc68, line 27: what about the case where the account doesn't exist? Shouldn't it return 404 instead of 500?

internal/adapters/entrypoints/rest/handlers/set_trusted_account.go
```diff
@@ -0,0 +1,55 @@
...
+	"github.com/rsksmart/liquidity-provider-server/pkg"
+)
+
+// NewSetTrustedAccountHandler
+// @Title Update Trusted Account
+// @Description Updates an existing trusted account
+// @Param TrustedAccountRequest body pkg.TrustedAccountRequest true "Details of the trusted account to update"
+// @Success 204 object
+// @Route /management/trusted-accounts [post]
+func NewSetTrustedAccountHandler(useCase *lpuc.SetTrustedAccountUseCase) http.HandlerFunc {
```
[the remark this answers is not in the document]
Luisfc68: Same comment in this iteration. Both "Set" and "Add" terms can be used for data insertion, so it is confusing, the naming in this scenario should be "Update" rather than "Set"

internal/adapters/entrypoints/rest/handlers/set_trusted_account.go
```diff
@@ -0,0 +1,42 @@
...
+			return
+		}
+		accountDetails := lp.TrustedAccountDetails{
+			Address:          request.Address,
+			Name:             request.Name,
+			Btc_locking_cap:  entities.NewBigWei(request.BtcLockingCap),
+			Rbtc_locking_cap: entities.NewBigWei(request.RbtcLockingCap),
+		}
+		err = useCase.Run(req.Context(), accountDetails)
+		if err != nil {
```
Luisfc68, line 35: same comment here, you're returning 500 in all the cases without worrying about the error

internal/configuration/registry/usecase.go
```diff
@@ -58,6 +59,11 @@ type UseCaseRegistry struct {
 	availableLiquidityUseCase     *liquidity_provider.GetAvailableLiquidityUseCase
 	updatePeginDepositUseCase     *watcher.UpdatePeginDepositUseCase
 	getServerInfoUseCase          *liquidity_provider.ServerInfoUseCase
+	trustedAccountRepository      entities_lp.TrustedAccountRepository
```
Luisfc68, line 62: this is the use case registry, this shouldn't be here, there is a database registry

internal/configuration/registry/usecase.go
```diff
@@ -233,6 +239,23 @@ func NewUseCaseRegistry(
 		availableLiquidityUseCase: liquidity_provider.NewGetAvailableLiquidityUseCase(liquidityProvider, liquidityProvider, liquidityProvider),
 		updatePeginDepositUseCase: watcher.NewUpdatePeginDepositUseCase(databaseRegistry.PeginRepository),
 		getServerInfoUseCase:      liquidity_provider.NewServerInfoUseCase(),
+		trustedAccountRepository:  databaseRegistry.TrustedAccountRepository,
```
Luisfc68, line 242: same here

internal/usecases/common.go
```diff
@@ -48,6 +49,7 @@ const (
 	SetPeginConfigId           UseCaseId = "SetPeginConfigUseCase"
 	SetPegoutConfigId          UseCaseId = "SetPegoutConfigUseCase"
 	SetGeneralConfigId         UseCaseId = "SetGeneralConfigUseCase"
+	SetTrustedAccountId        UseCaseId = "SetTrustedAccountUseCase"
```
Luisfc68, line 52: you're only adding `SetTrustedAccountId`, you're missing the other use cases

internal/usecases/liquidity_provider/add_trusted_account.go
```diff
@@ -0,0 +1,39 @@
...
+) *AddTrustedAccountUseCase {
+	return &AddTrustedAccountUseCase{
+		trustedAccountRepository: trustedAccountRepository,
+		signer:                   signer,
+		hashFunc:                 hashFunc,
+	}
+}
+
+func (useCase *AddTrustedAccountUseCase) Run(ctx context.Context, account liquidity_provider.TrustedAccountDetails) error {
+	_, err := usecases.SignConfiguration(usecases.SetTrustedAccountId, useCase.signer, useCase.hashFunc, account)
```
Luisfc68, line 30: Why are you signing and then not using the result? We discussed this during the refinement of the epic, without the signature anyone can manipulate the trusted accounts directly from the DB. You need to store the signed version and check the integrity on every read.

internal/usecases/liquidity_provider/set_trusted_account.go
```diff
@@ -0,0 +1,39 @@
...
+) *SetTrustedAccountUseCase {
+	return &SetTrustedAccountUseCase{
+		trustedAccountRepository: trustedAccountRepository,
+		signer:                   signer,
+		hashFunc:                 hashFunc,
+	}
+}
+
+func (useCase *SetTrustedAccountUseCase) Run(ctx context.Context, account liquidity_provider.TrustedAccountDetails) error {
+	_, err := usecases.SignConfiguration(usecases.SetTrustedAccountId, useCase.signer, useCase.hashFunc, account)
```
Luisfc68, line 30: this is also not using the result of the signature, this use case is incomplete.

internal/adapters/dataproviders/database/mongo/trusted_account.go
```diff
@@ -0,0 +1,107 @@
...
+	logDbInteraction(Insert, account)
+	return nil
+}
+
+func (repo *trustedAccountMongoRepository) UpdateTrustedAccount(ctx context.Context, account liquidity_provider.TrustedAccountDetails) error {
+	dbCtx, cancel := context.WithTimeout(ctx, repo.conn.timeout)
+	defer cancel()
+	collection := repo.conn.Collection(TrustedAccountCollection)
+	_, err := repo.GetTrustedAccount(ctx, account.Address)
+	if err != nil && !errors.Is(err, liquidity_provider.ErrTrustedAccountNotFound) {
```
Luisfc68, line 79: in this case you want the account to exists, so if you encounter `ErrTrustedAccountNotFound ` the error should be returned too

Luisfc68 (changes requested):
The handlers and pkg converters are not tested, and in the case of the handlers they are returning status code 500 for everything regardless of the specific error.

VERY IMPORTANT: 
The add use case doesn't sign the trusted accounts and none of the get uses cases checks the integrity of the TA read from the database (because they weren't signed at first place). This cannot be ignored because in the same way as we do with the server configuration, we need to prevent those values to be modified in case someone gains access to the DB. To summarize the PR is missing functionality.

internal/usecases/liquidity_provider/get_trusted_accounts.go
```diff
@@ -0,0 +1,23 @@
...
+func NewGetTrustedAccountsUseCase(
+	trustedAccountRepository liquidity_provider.TrustedAccountRepository,
+) *GetTrustedAccountsUseCase {
+	return &GetTrustedAccountsUseCase{
+		trustedAccountRepository: trustedAccountRepository,
+	}
+}
+
+func (useCase *GetTrustedAccountsUseCase) Run(ctx context.Context) ([]liquidity_provider.TrustedAccountDetails, error) {
+	return useCase.trustedAccountRepository.GetAllTrustedAccounts(ctx)
```
Luisfc68, line 22: I think you should verify the integrity of the TA here and return an error if its tampered

internal/usecases/liquidity_provider/set_trusted_account.go
Luisfc68, line 1: like in the handlers this file also needs to be renamed to update instead of set

internal/usecases/liquidity_provider/get_trusted_accounts.go
Luisfc68: with its tests of course

internal/adapters/entrypoints/rest/assets/static/management.js
```diff
@@ -571,6 +572,145 @@ const addCollateral = async (amountId, endpoint, elementId, loadingBarId, button
...
+    try {
+        const response = await fetch('/management/trusted-accounts', {
+            method: 'GET',
+            headers: {
+                'Content-Type': 'application/json',
+                'X-CSRF-Token': csrfToken
+            }
+        });
+        const data = await response.json();
+        console.log(data);
```
Luisfc68, line 587:
```suggestion
```

internal/adapters/entrypoints/rest/assets/static/management.js
```diff
@@ -571,6 +572,145 @@ const addCollateral = async (amountId, endpoint, elementId, loadingBarId, button
...
+    const name = document.getElementById('accountName').value.trim();
+    const address = document.getElementById('accountAddress').value.trim();
+    let btcLockingCap = document.getElementById('btc_locking_cap').value.trim();
+    let rbtcLockingCap = document.getElementById('rbtc_locking_cap').value.trim();
+    if (!name || !address) {
+        showErrorToast('Account name and address are required.');
+        return;
+    }
+    try {
+        btcLockingCap = btcLockingCap ? new Decimal(etherToWei(btcLockingCap)).toNumber() : 0;
```
Luisfc68, line 657: use bigint here to prevent overflow

Luisfc68 (changes requested):
PR is missing the integrity check when fetching the TAs and some tests, besides that the rest looks fine

MaximStanciu8 (approved):

internal/adapters/entrypoints/rest/assets/static/management.js
Dominikkq: Then it becomes a string and I need a number in request, using this Decimal method should already prevent any overflow

Dominikkq: DecimalJS should already prevent overflow, and would have to be converted back to number(with probably more unsafe methods) when submitting request body for json stringify

Luisfc68 (approved):
LGTM, remember to replace `IntegrityCheck` by `ValidateConfiguration` for the check of the trusted accounts once @AndresQuijano PRs is merged

AndresQuijano (approved):

PR #704, opened 2025-05-22
feat: Replace integrity check

Dominikkq:
**What:**
Replace `CheckIntegrity` with `ValidateConfiguration`
Update tests

### DIFF
--- a/internal/adapters/entrypoints/rest/handlers/get_trusted_accounts_test.go
+++ b/internal/adapters/entrypoints/rest/handlers/get_trusted_accounts_test.go
@@ -54,7 +54,9 @@ func TestNewGetTrustedAccountsHandler(t *testing.T) {
 		repo.On("GetAllTrustedAccounts", mock.Anything).Return(mockSignedAccounts, nil)
 		hashMock := &mocks.HashMock{}
 		hashMock.On("Hash", mock.Anything).Return(mockHashBytes)
-		useCase := lpuc.NewGetTrustedAccountsUseCase(repo, hashMock.Hash)
+		signerMock := &mocks.SignerMock{}
+		signerMock.On("Validate", mock.Anything, mock.Anything).Return(true)
+		useCase := lpuc.NewGetTrustedAccountsUseCase(repo, hashMock.Hash, signerMock)
 		handler := http.HandlerFunc(handlers.NewGetTrustedAccountsHandler(useCase))
 		handler.ServeHTTP(recorder, request)
 		assert.Equal(t, http.StatusOK, recorder.Code)
@@ -68,6 +70,7 @@ func TestNewGetTrustedAccountsHandler(t *testing.T) {
 		assert.Equal(t, "Test Account 2", response.Accounts[1].Name)
 		repo.AssertExpectations(t)
 		hashMock.AssertExpectations(t)
+		signerMock.AssertExpectations(t)
 	})
 	t.Run("should return 500 on error", func(t *testing.T) {
 		recorder := httptest.NewRecorder()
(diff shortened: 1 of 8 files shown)


MaximStanciu8 (approved):

internal/entities/liquidity_provider/liquidity_provider.go
```diff
@@ -134,15 +136,15 @@ func ValidateConfiguration[T ConfigurationType](
...
+		return nil, ErrConfigurationNotFound
 	}
 	if err = configuration.CheckIntegrity(hashFunction); err != nil {
 		log.Errorf("Tampered %s configuration. Using default configuration. Error: %v", displayName, err)
 		return nil, err
 	}
 	if !signer.Validate(configuration.Signature, configuration.Hash) {
 		log.Errorf("Invalid %s configuration signature. Using default configuration.", displayName)
-		return nil, errors.New("invalid signature")
+		return nil, ErrInvalidSignature
```
AndresQuijano, line 147: I like this. It's something that was missing in my previous PR. Can you please add also another sentinel error for the case of the tampered configuration and then remove all the logs? This is an entity and should not be showing any log.

internal/adapters/entrypoints/rest/handlers/get_trusted_accounts_test.go
```diff
@@ -68,18 +70,21 @@ func TestNewGetTrustedAccountsHandler(t *testing.T) {
...
 		hashMock := &mocks.HashMock{}
-		useCase := lpuc.NewGetTrustedAccountsUseCase(repo, hashMock.Hash)
+		signerMock := &mocks.SignerMock{}
+		useCase := lpuc.NewGetTrustedAccountsUseCase(repo, hashMock.Hash, signerMock)
 		handler := http.HandlerFunc(handlers.NewGetTrustedAccountsHandler(useCase))
 		handler.ServeHTTP(recorder, request)
 		assert.Equal(t, http.StatusInternalServerError, recorder.Code)
 		repo.AssertExpectations(t)
 		hashMock.AssertNotCalled(t, "Hash")
+		signerMock.AssertNotCalled(t, "Validate")
```
AndresQuijano, line 88: I just noticed that the handler under test here is not handling all possible errors and is mapping all of them to a generic 500 one. Is this on purpose? Otherwise I think we should manage each error and sent a response based on the type of error and write tests accordingly.

AndresQuijano (approved):

internal/usecases/liquidity_provider/get_trusted_account.go
```diff
@@ -10,27 +11,42 @@ import (
...
 	}
-	if err := signedAccount.CheckIntegrity(useCase.hashFunction); err != nil {
+	signedAccount, err := liquidity_provider.ValidateConfiguration(
+		"trusted account",
+		useCase.signer,
+		readFunction,
+		useCase.hashFunction,
+	)
+	if err != nil {
+		if errors.Is(err, liquidity_provider.ErrTrustedAccountNotFound) || errors.Is(err, liquidity_provider.ErrConfigurationNotFound) {
```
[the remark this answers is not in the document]
AndresQuijano: Updated

internal/usecases/liquidity_provider/get_trusted_account.go
```diff
@@ -10,28 +11,45 @@ import (
...
+		return signedAccount, nil
 	}
-	if err := signedAccount.CheckIntegrity(useCase.hashFunction); err != nil {
-		return nil, liquidity_provider.ErrTamperedTrustedAccount
+	signedAccount, err := liquidity_provider.ValidateConfiguration(
+		useCase.signer,
+		readFunction,
+		useCase.hashFunction,
+	)
+	if err != nil {
```
Luisfc68, line 45: errors.Is returns false if the err is nil, there is no need of checking first if error != nil, you can skip that to avoid the nesting and have something like:
```
if errors.Is(err, liquidity_provider.TrustedAccountNotFoundError) {
    return nil, err
} else if errors.Is(err, liquidity_provider.ConfigurationNotFoundError) {
    return nil, liquidity_provider.TrustedAccountNotFoundError
} else if err != nil {
    return nil, liquidity_provider.TamperedTrustedAccountError
}
```

image

Luisfc68 (approved):

PR #709, opened 2025-05-29
feat: added the asset report endpoint

gsoares85:
# What
Added the asset report endpoint
https://rsklabs.atlassian.net/browse/GBI-2540

### DIFF
--- a/OpenApi.yml
+++ b/OpenApi.yml
@@ -521,7 +521,6 @@ components:
       properties:
         fee:
           type: integer
-          deprecated: true
         feePercentage:
           type: number
         fixedFee:
@@ -1034,6 +1033,13 @@ paths:
         "204":
           description: ""
       summary: Withdraw PegIn Collateral
+  /reports/assets:
+    get:
+      description: ' Get the asset information for the LPS.'
+      responses:
+        "200":
+          description: ""
+      summary: Get asset Reports
   /reports/pegin:
     get:
       description: ' Get the last pegins on the API. Included in the management API.'
--- a/internal/adapters/entrypoints/rest/handlers/get_reports_assets.go
+++ b/internal/adapters/entrypoints/rest/handlers/get_reports_assets.go
@@ -1,0 +1,25 @@
+package handlers
+
+import (
+	"github.com/rsksmart/liquidity-provider-server/internal/adapters/entrypoints/rest"
+	"github.com/rsksmart/liquidity-provider-server/internal/usecases/reports"
+	"net/http"
+)
+
+// NewGetReportsAssetsHandler
+// @Title Get asset Reports
+// @Description Get the asset information for the LPS.
+// @Success 200 pkg.GetAssetsReportResponse
+// @Route /reports/assets [get]
+func NewGetReportsAssetsHandler(useCase *reports.GetAssetsReportUseCase) http.HandlerFunc {
+	return func(w http.ResponseWriter, req *http.Request) {
+		response, err := useCase.Run(req.Context())
+		if err != nil {
+			jsonErr := rest.NewErrorResponseWithDetails("Request error", rest.DetailsFromError(err), false)
+			rest.JsonErrorResponse(w, http.StatusBadRequest, jsonErr)
+			return
+		}
+
+		rest.JsonResponseWithBody(w, http.StatusOK, &response)
+	}
+}
--- a/internal/usecases/reports/get_assets_report.go
+++ b/internal/usecases/reports/get_assets_report.go
@@ -1,0 +1,153 @@
+package reports
+
+import (
+	"context"
+	"github.com/rsksmart/liquidity-provider-server/internal/entities"
+	"github.com/rsksmart/liquidity-provider-server/internal/entities/blockchain"
+	"github.com/rsksmart/liquidity-provider-server/internal/entities/liquidity_provider"
+	"github.com/rsksmart/liquidity-provider-server/internal/entities/quote"
+	"github.com/rsksmart/liquidity-provider-server/pkg"
+)
+
+type GetAssetsReportUseCase struct {
+	btcWallet        blockchain.BitcoinWallet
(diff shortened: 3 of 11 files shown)


OpenApi.yml
```diff
@@ -521,7 +521,6 @@ components:
       properties:
         fee:
           type: integer
-          deprecated: true
```
Luisfc68, line 524: please remember we're maintaining this parameter manually, don't remove it

pkg/liquidity_provider.go
```diff
@@ -152,6 +152,15 @@ type GetRevenueReportResponse struct {
 	TotalProfit        *big.Int `json:"total_profit" validate:"required"`
 }
 
+type GetAssetsReportResponse struct {
```
Luisfc68, line 155: we use camelCase in the API fields not snake_case

internal/adapters/entrypoints/rest/handlers/get_reports_assets.go
Luisfc68, line 1: not tested

internal/usecases/reports/get_assets_report.go
```diff
@@ -0,0 +1,153 @@
+package reports
+
+import (
+	"context"
+	"github.com/rsksmart/liquidity-provider-server/internal/entities"
+	"github.com/rsksmart/liquidity-provider-server/internal/entities/blockchain"
+	"github.com/rsksmart/liquidity-provider-server/internal/entities/liquidity_provider"
+	"github.com/rsksmart/liquidity-provider-server/internal/entities/quote"
+	"github.com/rsksmart/liquidity-provider-server/pkg"
```
Luisfc68, line 9: this import is from an external layer, remember you should only import things from the same layer or inner ones

internal/usecases/reports/get_assets_report.go
```diff
@@ -0,0 +1,153 @@
...
+		rsk:              rsk,
+		lp:               lp,
+		peginProvider:    peginProvider,
+		pegoutProvider:   pegoutProvider,
+		peginRepository:  peginRepository,
+		pegoutRepository: pegoutRepository,
+	}
+}
+
+func (useCase *GetAssetsReportUseCase) Run(ctx context.Context) (pkg.GetAssetsReportResponse, error) {
```
Luisfc68, line 42: don't return a DTO in the use case, return a domain object and then convert it to DTO in the handler

internal/usecases/reports/get_assets_report.go
```diff
@@ -0,0 +1,153 @@
...
+	for _, retainedQuote := range quotes {
+		lockedPegout.Add(lockedPegout, retainedQuote.RequiredLiquidity)
+	}
+
+	return lockedPegout, nil
+}
+
+func (useCase *GetAssetsReportUseCase) GetRBTCLocked(ctx context.Context) (*entities.Wei, error) {
+	lockedPegin := entities.NewWei(0)
+	peginQuotes, err := useCase.peginRepository.GetRetainedQuoteByState(ctx, quote.PeginStateWaitingForDeposit)
```
Luisfc68, line 121: what about waiting for deposit confirmations?

internal/usecases/reports/get_assets_report.go
```diff
@@ -0,0 +1,153 @@
...
+	if err != nil {
+		return nil, err
+	}
+	for _, retainedQuote := range peginQuotes {
+		lockedPegin.Add(lockedPegin, retainedQuote.RequiredLiquidity)
+	}
+	pegoutQuotes, err := useCase.pegoutRepository.GetRetainedQuoteByState(ctx, quote.PegoutStateRefundPegOutSucceeded)
```
Luisfc68, line 128: this doesn't count as locked, it is true that is not counted as liquidity neither, but we're not saving those RBTC for any user, they are just waiting for rebalance

internal/usecases/reports/get_assets_report_test.go
```diff
@@ -0,0 +1,209 @@
...
+	require.Equal(t, result.BtcLocked, big.NewInt(0))
+}
+
+func TestGetAssetsReportUseCase_Run_rbtcBalanceError(t *testing.T) {
+	ctx := context.Background()
+
+	rskAddress := "rskAddress"
+
+	wallet := mocks.NewBitcoinWalletMock(t)
+	wallet.On("GetBalance").Return(nil, assert.AnError)
```
Luisfc68, line 147: if you return the error when checking the btc balance you won't reach the point where you check the rbtc balance

internal/usecases/reports/get_assets_report_test.go
Luisfc68, line 1: there are missing multiple scenarios, you're just checking three of them, get btc balance, get rbtc balance and reading the pegin quotes from the db

Luisfc68 (changes requested):

internal/usecases/reports/get_assets_report_test.go
```diff
@@ -0,0 +1,357 @@
...
+	result, err := useCase.Run(ctx)
+
+	require.Error(t, err)
+	require.Equal(t, result.RbtcLiquidity, big.NewInt(0))
+	require.Equal(t, result.BtcLiquidity, big.NewInt(0))
+	require.Equal(t, result.BtcBalance, big.NewInt(0))
+	require.Equal(t, result.RbtcBalance, big.NewInt(0))
+	require.Equal(t, result.RbtcLocked, big.NewInt(0))
+	require.Equal(t, result.BtcLocked, big.NewInt(0))
+	lp.AssertExpectations(t)
```
Luisfc68, line 144: you only mocked the execution of GetBalance on the btc wallet, instead of `AssertExpectations` you should use `AssertNotCalled` for the ones you didn't set up the mock because the function is supposed to return earlier, the same happens with the other error tests of this file

Luisfc68 (approved):

PR #716, opened 2025-06-04
fix: improve error message format/clarity | GBI-2642

Dominikkq:
**What:**
Rewrote error messages for percentageFee to further improve clarity

**Task:**
[GBI-2642](https://rsklabs.atlassian.net/browse/GBI-2642)

### DIFF
--- a/internal/adapters/entrypoints/rest/common.go
+++ b/internal/adapters/entrypoints/rest/common.go
@@ -135,6 +135,19 @@ func DecodeRequest[T any](w http.ResponseWriter, req *http.Request, body *T) err
 }
 
 func getValidationMessage(field validator.FieldError) string {
+	if field.Field() == "FeePercentage" {
+		switch field.Tag() {
+		case "gte":
+			if field.Param() == "0" {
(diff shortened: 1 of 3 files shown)


internal/adapters/entrypoints/rest/common.go
```diff
@@ -135,17 +135,27 @@ func DecodeRequest[T any](w http.ResponseWriter, req *http.Request, body *T) err
 }
 
 func getValidationMessage(field validator.FieldError) string {
+	if field.Field() == "FeePercentage" {
```
Luisfc68, line 138: I think is better to have this kind of handling of the error msg in the front. In there you have the context of which field is being manipulated and can provide the proper msg on each case. If you go this way this function will become huge the more specific you want to be in the error msg. If any, maybe you can return something similar to the previous version to indicate the field and the rule being broken, but this level of responsibility is too much for this function

Luisfc68 (changes requested):

MaximStanciu8 (approved):

PR #730, opened 2025-06-10
Feature/gbi 2681 store gas used

gsoares85:
# What

Stored gas costs for pegin and pegouts operations.

[https://rsklabs.atlassian.net/browse/GBI-2681](https://rsklabs.atlassian.net/browse/GBI-2681)


internal/adapters/dataproviders/rootstock/lbc.go
Luisfc68, line 1: all the updated function contain the same error related to the fee and the gas

Luisfc68 (changes requested):
This PR is mixing the concept of gas used with the fee paid for a transaction. If the purpose of the task is to store the information needed to calculate the profit later, then storing the gas used is not enough, also it shouldn't be stored as wei, because gas is not a currency, the currency of the fee of the tx stills being wei and is the result of multiplying the gas used by the gas price for that tx. The PR needs to be modified taking into account the difference between the gas and the gas price

internal/adapters/dataproviders/rootstock/wallet.go
```diff
@@ -92,21 +98,23 @@ func (wallet *RskWalletImpl) SendRbtc(ctx context.Context, config blockchain.Tra
...
 	if err != nil {
-		return "", err
+		return receiptData, err
 	} else if receipt == nil || receipt.Status == 0 {
-		return txHash, fmt.Errorf("%s transaction failed", txHash)
+		return receiptData, fmt.Errorf("%s transaction failed", receiptData.TransactionHash)
 	}
-	return signedTx.Hash().String(), nil
+	receiptData.GasUsed = big.NewInt(int64(receipt.GasUsed))
+	return receiptData, nil
```
Dominikkq, line 117: GasPrice always 0 (it doensn't seem to be copied from receipt.EffectiveGasPrice).
GasUsed built with int64(receipt.GasUsed), which can technically overflow(but shouldn't realistacally so probably optional)

Dominikkq (changes requested):

Luisfc68 (changes requested):
Please, try to prefer 
```
n := new(big.Int)
n.SetUint64(xxx)
```
instead of using `big.NewInt` and casting the argument, because we've been doing some work removing the risks of overflow/underflow and that way of creating big.Int has some risks

internal/adapters/dataproviders/rootstock/bridge_test.go
```diff
@@ -260,7 +260,7 @@ func TestRskBridgeImpl_RegisterBtcCoinbaseTransaction(t *testing.T) {
 	t.Run("Should handle error when tx reverts", func(t *testing.T) {
 		bridgeMock := &mocks.RskBridgeBindingMock{}
 		mockClient := &mocks.RpcClientBindingMock{}
-		tx := prepareTxMocks(mockClient, signerMock, false)
+		tx, _ := prepareTxMocks(mockClient, signerMock, false)
```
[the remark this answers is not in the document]
gsoares85: In this test the receipt is not needed, but there are other tests like this one
[feature/GBI-2681-storeGasUsed](https://github.com/rsksmart/liquidity-provider-server/tree/feature/GBI-2681-storeGasUsed)
Where the receipt is used.

Luisfc68 (changes requested):
The TransactionReceipt struct is very big and I see a lot of times in the code the pattern of creating it and populating each field with the zero value manually. This will cause that if we add a new field later we'll need to touch every place where the empty receipt is initialized, I recommend implementing a NewEmptyReceipt() function and replace the manual initialization with it so the code is less polluted with the initialization of the empty struct every time a error is handled

PR #736, opened 2025-06-16
Feature/GBI-2687 - Eclipse attack detection

Luisfc68:
## What
Implementation of the Eclipse Attack improvement proposal made by security team. The PR includes:
- Moving the ticker wrapper to the utils package
- Adding the eclipse check use case and handler
- Added extra datasources array to the registry

## Why
To be able to detect isolation of one of the nodes used by the LPS

## Task
https://rsklabs.atlassian.net/browse/GBI-2687


internal/usecases/watcher/eclipse_check_test.go
```diff
@@ -0,0 +1,581 @@
...
+		rskExtras := []*mocks.RootstockRpcServerMock{rskExtra1, rskExtra2}
+		rskRpc.EXPECT().GetBlockByNumber(mock.Anything, nilBigInt).Return(blockchain.BlockInfo{
+			Hash:      "otherHash",
+			Number:    300,
+			Timestamp: timestamp,
+			Nonce:     nonce,
+		}, nil).Once()
+		rskRpc.EXPECT().GetBlockByNumber(mock.Anything, nilBigInt).Return(blockchain.BlockInfo{
+			Hash:      test.AnyHash,
+			Number:    500,
```
AndresQuijano, line 165: I suggest using `blocks` here instead of 500 for clarity

internal/usecases/watcher/eclipse_check_test.go
```diff
@@ -0,0 +1,581 @@
...
+		require.NoError(t, err)
+		btcRpc.AssertExpectations(t)
+		for _, btcExtra := range btcExtras {
+			btcExtra.AssertExpectations(t)
+		}
+		eventBus.AssertNotCalled(t, "Publish")
+		alertSender.AssertNotCalled(t, "SendAlert")
+		mutex.AssertExpectations(t)
+	})
+	t.Run("should trigger the alert if eclipse attack is detected", func(t *testing.T) {
```
AndresQuijano, line 280: This test also use the explicit values of blocks and headers instead of the constants

AndresQuijano (dismissed):

PR #751, opened 2025-07-09
feat: GBI-2563 | additional validation

Dominikkq:
**What:**
Check the minPegin is above the bridge minimum in SetPeginConfigUseCase.
Validate all the wei values are positive in the three types of configuration.
Check that both rsk and btc ConfirmationsPerAmount have at least one entry in SetGeneralConfigUseCase , and that the keys of the map are all positive.

**Task:**
[GBI-2563](https://rsklabs.atlassian.net/browse/GBI-2563)


alexjavabraz (approved):
LGTM

Luisfc68 (changes requested):
This PR should be directed to the v2.3.0 branch to be integrate with the changes in that version (data type changes)

PR #765, opened 2025-07-23
feat: track btc release for lps pegouts

Luisfc68:
## What
Add a new watcher to check every time the `batch_pegout_created` event is emmited by the bridge. If the batch contains any of the native pegouts made by the LPS to rebalance, then the LPS will mark those quotes with a new state called `BtcReleased`

## Why
For monitoring purposes and the financial API, we need to be able to distinguish between the rebalance pegouts that are already completed an the ones in progress

## Important
This PR introduces 3 new environment variables; `BTC_RELEASE_WATCHER_START_BLOCK`, `BTC_RELEASE_WATCHER_PAGE_SIZE`, and `BTC_RELEASE_CHECK_TIMEOUT`. This is important because this is how we're going to update the existing quotes in the DB. Instead of updating them manually or developing an endpoint specifically to add the new field to the quotes in the DB, it is simpler to use this env vars to make the LPS update the quotes itself. If `BTC_RELEASE_WATCHER_START_BLOCK` is zero, then the LPS will start looking from the latest block, instead of that we might one the LP registration block as start block and a relatively big page size so the LPS can update all the quotes across that LP history, depending on the network this might take a couple days, but after that all the quotes in the DB should be up to date and the LPS could start with `BTC_RELEASE_WATCHER_START_BLOCK=0`

## Task
https://rsklabs.atlassian.net/browse/GBI-2791

### DIFF
--- a/internal/adapters/entrypoints/watcher/btc_release_watcher.go
+++ b/internal/adapters/entrypoints/watcher/btc_release_watcher.go
@@ -126,12 +126,12 @@ func (watcher *BtcReleaseWatcher) checkBatchPegOuts() (uint64, error) {
 }
 
 func (watcher *BtcReleaseWatcher) nextBlock(ctx context.Context) (uint64, error) {
+	var toBlock uint64 = 0
 	height, err := watcher.rpc.Rsk.GetHeight(ctx)
 	if err != nil {
-		return 0, fmt.Errorf("error getting RSK height in BtcReleaseWatcher: %w", err)
+		return toBlock, fmt.Errorf("error getting RSK height in BtcReleaseWatcher: %w", err)
 	}
 
-	var toBlock uint64
 	if (watcher.currentBlock + watcher.pageSize) < height {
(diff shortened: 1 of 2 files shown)


internal/adapters/entrypoints/watcher/btc_release_watcher.go
```diff
@@ -0,0 +1,163 @@
...
+		return 0, err
+	}
+
+	return toBlock, nil
+}
+
+func (watcher *BtcReleaseWatcher) nextBlock(ctx context.Context) (uint64, error) {
+	height, err := watcher.rpc.Rsk.GetHeight(ctx)
+	if err != nil {
+		return 0, fmt.Errorf("error getting RSK height in BtcReleaseWatcher: %w", err)
```
alexjavabraz, line 131:
```suggestion
		return toBlock, fmt.Errorf("error getting RSK height in BtcReleaseWatcher: %w", err)
```

alexjavabraz (approved):
LGTM

PR #821, opened 2025-09-17
Feature/GBI-2869 - LBC Split implementation

Luisfc68:
## What
- Fix and update integration tests
- Add new binding script
- Implement LBC Split in the LPS

## Why
Because the initial iteration of the LBC split has been done in the contract side. Now we're moving to the adaptations required in the server. This only includes adapting to the new interfaces, doesn't include any change in the server behavior.

## Important
This PR only contains changes in the integration tests and the source code, it **does not** include changes in the unit test, therefore, expect the pipeline to fail on this PR. The reason of this is that I'm including the unit test in a separate PR to reduce the size of the code being reviewed.

Additionally, we're merging all these intermediate PRs to the `lbc-split` branch. This is the feature branch that later will be merged into the version branch.

## Task
https://rsklabs.atlassian.net/browse/GBI-2869


AndresQuijano:
My only observation on this PR is the naming convention. Sometimes it is pegIn and sometimes pegin. Same for pegout/pegOut.
I noticed an issue in /internal/adapters/dataproviders/rootstock/collateral_management_contract.go but I see it fixed in the next PR.

Luisfc68:
@AndresQuijano Thanks, I see your point, we're using pegIn in the contract and pegin here, therefore the bindings are also with pegIn. Since changing the whole project to use pegIn as well is too expensive I propose to leave like that (pegIn -> contract, pegin -> LPS). If you find any pegIn outside of the bindings, then I think we should change that one particularly

PR #828, opened 2025-09-24
fix: fetch retiring federation info

Luisfc68:
## What
Add a new env var to optionally use the retiring federation

## Why
This is a useful option to have in the server to test certain scenarios. It should not be used in production


internal/adapters/dataproviders/rootstock/federation/federation.go
```diff
@@ -271,6 +272,7 @@ func CalculateFlyoverDerivationAddress(fedInfo rootstock.FederationInfo, btcPara
 	if fedInfo.UseSegwit {
 		addressScriptHash, err = bitcoin.ScriptToAddressP2shP2wsh(flyoverScript, &btcParams)
 	} else {
+		log.Warn("Using non-segwit federation. Consider changing to a segwit compatible federation if its supported by the network.")
```
AndresQuijano, line 275: We should add a test to check that this is actually working

AndresQuijano (approved):

PR #829, opened 2025-09-25
test: adapt unit test to lbc split

Luisfc68:
## What
Adapt unit test to LBC split implementation 

## Why
This PR was separated from the one touching the source code (#821) to reduce the size of that PR so the reviewers could focus more in the code changes

## Task
https://rsklabs.atlassian.net/browse/FLY-570


internal/adapters/dataproviders/rootstock/abi.go
```diff
@@ -0,0 +1,51 @@
...
+	}
+	pegOutAbi, err := bindings.IPegOutMetaData.GetAbi()
+	if err != nil {
+		panic("could not load PegOut ABI: " + err.Error())
+	}
+	discoveryAbi, err := bindings.IFlyoverDiscoveryMetaData.GetAbi()
+	if err != nil {
+		panic("could not load Discovery ABI: " + err.Error())
+	}
+	collateralManagementAbi, err := bindings.IPegOutMetaData.GetAbi()
```
AndresQuijano, line 30: I think there's an error here. The abi is getting obtained from the IPegOutMetaData. I think it should be
bindings.ICollateralManagementMetaData.GetAbi()
This is probably a copy paste error

PR #891, opened 2026-01-08
Feature/fly 2162

Hakob23:
## What
Implement the excess tolerance in the Management UI

## Why
Add excess tolerance configuration

## Type of Change
- [ ] Test updates

- [ ] Management UI / API

## Related Issues
[Jira ticket](https://rsklabs.atlassian.net/jira/software/c/projects/FLY/boards/682?selectedIssue=FLY-2162)

## How to test
```go test -v ./internal/adapters/entrypoints/rest/handlers/... -run TestSetGeneralConfigHandler

go test -v ./internal/entities/liquidity_provider/... -run TestDefaultGeneralConfiguration

go test -v ./internal/entities/liquidity_provider/... -run TestValidateConfiguration

go test -v ./internal/entities/... -run TestSigned_CheckIntegrity

go test -v ./pkg/... -run TestFromGeneralConfigurationDTO

go test -v ./internal/adapters/dataproviders/database/mongo/... -run "TestLpMongoRepository_GetGeneralConfiguration|TestLpMongoRepository_UpsertGeneralConfiguration"

go test -v ./internal/adapters/dataproviders/... -run TestLocalLiquidityProvider_GeneralConfiguration
```

image
image

### DIFF
--- a/internal/adapters/entrypoints/rest/assets/static/management.js
+++ b/internal/adapters/entrypoints/rest/assets/static/management.js
@@ -20,7 +20,14 @@ const generalChanged = { value: false };
 const peginChanged = { value: false };
 const pegoutChanged = { value: false };
 
-const setTextContent = (id, text) => document.getElementById(id).textContent = text;
+const setTextContent = (id, text) => {
+    const element = document.getElementById(id);
+    if (!element) {
+        console.warn(`Element with id "${id}" not found`);
+        return;
+    }
+    element.textContent = text;
+};
 
 const fetchData = async (url, elementId, csrfToken) => {
     try {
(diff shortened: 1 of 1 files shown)


internal/adapters/entrypoints/rest/assets/static/management.js
AndresQuijano: In my opinion, from the UX perspective we should not delete the content of the box when the toggle is pressed/changed. When there's something there (previously configured) and we change the toggle twice, the value is lost and we have to refresh the whole page to get it back.
This is a small thing and is only a plus.

AndresQuijano (changes requested):
Not related to this PR but I've noticed that the frontend is validating that only one value is sent, either fixed or percentage, but we don't have the same validation in the backend. I don't know if this is intentional or if this is a mistake. If this is a mistake we can create a bug in a separated task to fix the backend.

PR #894, opened 2026-01-16
Feature/2.5 script update

Hakob23:
## What
Temporary solution for working with Flyover 2.5

## Why
The structure of LBC has changed from 1 contract to four contracts, and the current config doesn't support the new structure. This is a temporary solution as it requires a connection with LBC repository in the same computer in order to use its deployment scripts and deploy the contracts(branch QA-Test), until something more stable is configured.

## Type of Change

- [ ] Utility scripts
- [ ] Configuration files

## How to test

1. Have the LBC repo and QA-Test branch(Up-to-date) in the same local machine, make sure that all dependencies are installed
2. Run the following commands
```
docker system prune -a -f --volumes 
rm -rf docker-compose/local/volumes 
cd docker-compose/local 
export LPS_STAGE=regtest
./lps-env.sh up
```

```curl http://localhost:8080/health```

### DIFF
--- a/docker-compose/lps/Dockerfile
+++ b/docker-compose/lps/Dockerfile
@@ -1,8 +1,4 @@
-ARG USE_PREBUILT=source
-ARG TARGETARCH=amd64
-
-# Builder stage for compiling from source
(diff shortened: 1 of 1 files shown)


docker-compose/local/docker-compose.lps.yml
Luisfc68: this is what I mean, instead of this we could do it through https://docs.docker.com/reference/compose-file/build/#platforms

PR #898, opened 2026-01-23
Feature/FLY-2150 - ABIGEN v2 migration

Luisfc68:
## What
Migration to abigen v2, this changes how the bindings are generated, consumed, and tested for all the smart contracts. This new version of the library contains several breaking changes, that is the reason of the size of the PR

## Why
New versions of geth use abigen v2, avoiding the migration would have started to limit us in some ways, for example, there are already some frictions with abigen v1 when the integration test try to listen to the node WS to monitor the emitted events

## Type of Change
- [ ] Breaking change 
- [ ] Documentation update
- [x] Refactoring (no functional changes, no api changes)
- [ ] Performance improvement
- [ ] Test updates
- [ ] Security fix

- [ ] Management UI / API
- [x] PegIn flow
- [x] PegOut flow
- [x] Utility scripts
- [ ] Configuration files
- [ ] Metrics and alerting

## Related Issues
[Jira ticket](https://rsklabs.atlassian.net/browse/FLY-2150)

## How to test
This ticket doesn't contain any new functionality nor API changes, however it is recommended a full regression as all the functions that involve a smart contract interaction have been affected.

## Reviewer Guidelines
Please review carefully, there wasn't any way of reducing the size of the PR without including a hybrid version with the two versions of abigen running, therefore this PR contains the full migration

### DIFF
--- a/internal/adapters/dataproviders/rootstock/pegout_contract.go
+++ b/internal/adapters/dataproviders/rootstock/pegout_contract.go
@@ -150,7 +150,7 @@ func (pegoutContract *pegoutContractImpl) HashPegoutQuoteEIP712(pegoutQuote quot
 			if dataErr != nil {
 				return [32]byte{}, dataErr
 			}
-			return bind.Call(pegoutContract.contract, &bind.CallOpts{}, callData, pegoutContract.binding.UnpackHashPegOutQuote)
+			return bind.Call(pegoutContract.contract, &bind.CallOpts{}, callData, pegoutContract.binding.UnpackHashPegOutQuoteEIP712)
 		})
 	if err != nil {
 		return [32]byte{}, err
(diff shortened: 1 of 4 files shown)


internal/adapters/dataproviders/rootstock/pegout_contract.go
```diff
@@ -132,7 +146,11 @@ func (pegoutContract *pegoutContractImpl) HashPegoutQuoteEIP712(pegoutQuote quot
 
 	result, err = rskRetry(pegoutContract.retryParams.Retries, pegoutContract.retryParams.Sleep,
 		func() ([32]byte, error) {
-			return pegoutContract.contract.HashPegOutQuoteEIP712(&opts, parsedQuote)
+			callData, dataErr := pegoutContract.binding.TryPackHashPegOutQuoteEIP712(parsedQuote)
+			if dataErr != nil {
+				return [32]byte{}, dataErr
+			}
+			return bind.Call(pegoutContract.contract, &bind.CallOpts{}, callData, pegoutContract.binding.UnpackHashPegOutQuote)
```
AndresQuijano: The unpack function here doesn't match the pack function.
The pack uses `TryPackHashPegOutQuoteEIP712` but the unpack uses `UnpackHashPegOutQuote`. It should be `UnpackHashPegOutQuoteEIP712`.

It works today because both functions return the same type (bytes32), but it could break if they change in the future.

Luisfc68: thanks, it was working because is the same output

PR #901, opened 2026-01-29
feat/FLY 2158/add transfer excess to cold wallet mechanism

AndresQuijano:
## What
This PR includes the mechanism to execute the transfer of the liquidity excess from the hot to the cold wallet. As part of that, it introduces a new config called state to track the last time a hot to cold transfer was executed.
Also as part of this PR two watchers were included:
- First one to periodically execute the use case
- Second one to monitor the events and create metrics related to cold to hot wallet transfers and their reasons (threshold reached or time forced).

## Why
To keep the liquidity under some values and move profits to a more safe place.

## Type of Change
- [ ] Breaking change 
- [ ] Documentation update
- [ ] Performance improvement
- [x] Test updates
- [ ] Security fix

- [ ] Management UI / API
- [ ] PegIn flow
- [ ] PegOut flow
- [ ] Utility scripts
- [ ] Configuration files
- [x] Metrics and alerting
- [x] Liquidity management

## Related Issues
[Jira ticket](https://rsklabs.atlassian.net/browse/FLY-2158)

## How to test
- Given the complexity of this feature, this is a recommendation to ease the process of testing. Read the full instructions before starting the test to have the full context.
- Start the LPS infrastructure in local without starting the LPS (just modify the script to not start the LPS)
- Start the LPS locally without using docker
- Find and follow the logs. Since it is running without docker, they should appear in the standard output, the console where the LPS was started.
- Take a look to the /metrics endpoint and check that these metrics are not present:
lps_cold_wallet_last_transfer_amount
lps_cold_wallet_transfers
- The by-default liquidity of the RSK and BTC hot wallets are way bigger than the target liquidity. By default the target liquidity is 1 (0.5 for each network), so it is expected that the first execution of the watcher + use case, triggers a transfer from cold to hot wallet on each network (one for BTC, one for RSK).
- The first execution is expected (by-default) to be 5 minutes after the LPS has been started. That value is hardcoded at internal/adapters/entrypoints/watcher/common.go and can be optionally adjusted to something more frequent like 1 minute for consecutive observations. To have this configuration applied, the LPS should be restarted.
- After the first execution the log should show a clear message pointing the transfers and the /metrics endpoint should show the mentioned cold wallet metrics.
- The second execution, and following, should show in the logs that nothing was transferred, since there is not excess.
- To test the forced-by-time transfer, the COLD_WALLET_FORCE_TRANSFER_AFTER_SECONDS parameter in the .sample.env file can be manipulated and set to something like 30 seconds (again, restart the LPS to have this applied). Then the excess should be set to something bigger than the 0.5 target but smaller than the threshold. By default the threshold is set to 20% = 0.1BTC per network. 
- One way to trigger this excess, is sending a small transfer to the BTC wallet of 0.01 using something like 

`alias bi='bitcoin-cli -rpcport=5555 -rpcuser=test -rpcpassword=test -rpcconnect=127.0.0.1'`
and then
`bi -rpcwallet=main -named sendtoaddress fee_rate=25 address=n1jGDaxCW6jemLZyd9wmDHddseZwEMV9C6 amount=0.01`
(change the address if required)
- Check the balance of the address

- And for RSK using something like 
`curl -s -X POST "http://127.0.0.1:4444" -H "Content-Type: application/json" -d '{"jsonrpc":"2.0","method":"eth_sendTransaction","params": [{"from": "0xcd2a3d9f938e13cd947ec05abc7fe734df8dd826", "to": "0x9D93929A9099be4355fC2389FbF253982F9dF47c", "value": "0x38D7EA4C68000"}],"id":1}' | jq`

- Then, the new watcher  + use case execution, should execute the transfers forced by time and not because it reached the threshold.

- Check the logs and /metrics for evidence of the execution.

- Repeat the process multiple times and check /metrics to see the progress and correctness. 

- Other things to monitor:
Check the flyover/liquidityProvider collection in Mongo. This is the object where the state is persisted and it was introduced in this PR. It contains timestamps of the last transfers from hot to cold wallet.
All the transfers must be worth the payment. Check this env variables: 
BTC_MIN_TRANSFER_FEE_MULTIPLIER=5
RBTC_MIN_TRANSFER_FEE_MULTIPLIER=100

## Screenshots 

image
image
image
image

### DIFF
--- a/internal/usecases/liquidity_provider/transfer_excess_to_cold_wallet.go
+++ b/internal/usecases/liquidity_provider/transfer_excess_to_cold_wallet.go
@@ -260,7 +260,7 @@ func (useCase *TransferExcessToColdWalletUseCase) calculateExcessForBothNetworks
 
 // publishBtcTransferEvent emits a BTC cold wallet transfer event to the event bus, distinguishing
 // between threshold-triggered and time-forced transfers.
-func (useCase *TransferExcessToColdWalletUseCase) publishBtcTransferEvent(txResult *internalTxResult, isTimeForcingTransfer bool) {
+func (useCase *TransferExcessToColdWalletUseCase) publishBtcTransferEvent(txResult internalTxResult, isTimeForcingTransfer bool) {
 	if isTimeForcingTransfer {
 		useCase.eventBus.Publish(cold_wallet.BtcTransferredDueToTimeForcingEvent{
 			Event:  entities.NewBaseEvent(cold_wallet.BtcTransferredDueToTimeForcingEventId),
@@ -317,7 +317,7 @@ func (useCase *TransferExcessToColdWalletUseCase) handleBtcTransfer(excess *enti
 
 // publishRskTransferEvent emits an RBTC cold wallet transfer event to the event bus, distinguishing
 // between threshold-triggered and time-forced transfers.
-func (useCase *TransferExcessToColdWalletUseCase) publishRskTransferEvent(txResult *internalTxResult, isTimeForcingTransfer bool) {
+func (useCase *TransferExcessToColdWalletUseCase) publishRskTransferEvent(txResult internalTxResult, isTimeForcingTransfer bool) {
 	if isTimeForcingTransfer {
 		useCase.eventBus.Publish(cold_wallet.RbtcTransferredDueToTimeForcingEvent{
 			Event:  entities.NewBaseEvent(cold_wallet.RbtcTransferredDueToTimeForcingEventId),
@@ -435,26 +435,26 @@ func (useCase *TransferExcessToColdWalletUseCase) calculateExcess(
 
 // executeBtcTransfer estimates the BTC transaction fee, verifies the transfer is economical
 // (amount >= fee * multiplier), and sends the funds to the cold wallet.
-func (useCase *TransferExcessToColdWalletUseCase) executeBtcTransfer(amount *entities.Wei) (*internalTxResult, error) {
+func (useCase *TransferExcessToColdWalletUseCase) executeBtcTransfer(amount *entities.Wei) (internalTxResult, error) {
 	coldBtcAddress := useCase.coldWallet.GetBtcAddress()
 
 	feeEstimation, err := useCase.btcWallet.EstimateTxFees(coldBtcAddress, amount)
 	if err != nil {
-		return nil, err
+		return internalTxResult{}, err
 	}
 
 	// Check if transfer is economical: amount >= fee * multiplier
 	minWorthwhileAmount := new(entities.Wei).Mul(feeEstimation.Value, entities.NewUWei(useCase.btcMinTransferFeeMultiplier))
 	if amount.Cmp(minWorthwhileAmount) < 0 {
-		return nil, TransferNotEconomicalError
+		return internalTxResult{}, TransferNotEconomicalError
 	}
 
 	txResult, err := useCase.btcWallet.Send(coldBtcAddress, amount)
 	if err != nil {
-		return nil, err
+		return internalTxResult{}, err
 	}
 
-	return &internalTxResult{
+	return internalTxResult{
 		TxHash: txResult.Hash,
 		Amount: amount,
 		Fee:    feeEstimation.Value,
@@ -463,12 +463,12 @@ func (useCase *TransferExcessToColdWalletUseCase) executeBtcTransfer(amount *ent
 
 // executeRskTransfer estimates gas costs, subtracts them from the amount, verifies the transfer is economical
 // (net amount >= gasCost * multiplier), and sends the RBTC to the cold wallet.
-func (useCase *TransferExcessToColdWalletUseCase) executeRskTransfer(ctx context.Context, amount *entities.Wei) (*internalTxResult, error) {
+func (useCase *TransferExcessToColdWalletUseCase) executeRskTransfer(ctx context.Context, amount *entities.Wei) (internalTxResult, error) {
 	coldRskAddress := useCase.coldWallet.GetRskAddress()
 
 	gasPrice, err := useCase.rpc.Rsk.GasPrice(ctx)
 	if err != nil {
-		return nil, err
+		return internalTxResult{}, err
 	}
 
 	gasCost := new(entities.Wei).Mul(gasPrice, entities.NewWei(SimpleTransferGasLimit))
@@ -479,19 +479,19 @@ func (useCase *TransferExcessToColdWalletUseCase) executeRskTransfer(ctx context
 	// Check if transfer is economical: amountToTransfer >= gasCost * multiplier
 	minWorthwhileAmount := new(entities.Wei).Mul(gasCost, entities.NewUWei(useCase.rbtcMinTransferFeeMultiplier))
 	if amountToTransfer.Cmp(minWorthwhileAmount) < 0 {
-		return nil, TransferNotEconomicalError
+		return internalTxResult{}, TransferNotEconomicalError
 	}
 
 	config := blockchain.NewTransactionConfig(amountToTransfer, SimpleTransferGasLimit, gasPrice)
 
 	receipt, err := useCase.rskWallet.SendRbtc(ctx, config, coldRskAddress)
 	if err != nil {
-		return nil, err
+		return internalTxResult{}, err
 	}
 
 	actualFee := new(entities.Wei).Mul(receipt.GasPrice, entities.NewUWei(receipt.GasUsed.Uint64()))
 
-	return &internalTxResult{
+	return internalTxResult{
 		TxHash: receipt.TransactionHash,
 		Amount: amountToTransfer,
 		Fee:    actualFee,


internal/adapters/entrypoints/rest/routes/routes_test.go
```diff
@@ -255,6 +256,7 @@ func setupRegistryMock(registryMock *mocks.UseCaseRegistryMock) {
 	registryMock.EXPECT().DeleteTrustedAccountUseCase().Return(&liquidity_provider.DeleteTrustedAccountUseCase{})
 	registryMock.EXPECT().RecommendedPegoutUseCase().Return(&pegout.RecommendedPegoutUseCase{})
 	registryMock.EXPECT().RecommendedPeginUseCase().Return(&pegin.RecommendedPeginUseCase{})
+	registryMock.EXPECT().TransferExcessToColdWalletUseCase().Return(&liquidity_provider.TransferExcessToColdWalletUseCase{})
```
Luisfc68: I think this is a leftover from when you used the test handler

internal/adapters/dataproviders/database/mongo/liquidity_provider.go
```diff
@@ -101,6 +102,22 @@ func (repo *lpMongoRepository) UpsertCredentials(ctx context.Context, credential
 	return upsertConfiguration(dbCtx, repo, configToStore, false)
 }
 
+func (repo *lpMongoRepository) GetStateConfiguration(ctx context.Context) (*entities.Signed[liquidity_provider.StateConfiguration], error) {
+	dbCtx, cancel := context.WithTimeout(ctx, repo.conn.timeout)
+	defer cancel()
+	return getConfiguration[liquidity_provider.StateConfiguration](dbCtx, repo, stateConfigId, false)
```
Luisfc68: I don't see any problem on using the verbose version here, actually, I think is better for debugging purposes

internal/usecases/liquidity_provider/transfer_excess_to_cold_wallet.go
```diff
@@ -0,0 +1,482 @@
...
+type TransferExcessToColdWalletUseCase struct {
+	peginProvider                liquidity_provider.PeginLiquidityProvider
+	pegoutProvider               liquidity_provider.PegoutLiquidityProvider
+	generalProvider              liquidity_provider.LiquidityProvider
+	lpRepository                 liquidity_provider.LiquidityProviderRepository
+	coldWallet                   cold_wallet.ColdWallet
+	btcWallet                    blockchain.BitcoinWallet
+	rskWallet                    blockchain.RootstockWallet
+	rpc                          blockchain.Rpc
+	rskWalletMutex               sync.Locker
```
Luisfc68: you need btcWalletMutex as well

internal/usecases/liquidity_provider/transfer_excess_to_cold_wallet.go
```diff
@@ -0,0 +1,482 @@
...
+	btcMinTransferFeeMultiplier  uint64
```
Luisfc68: Maybe you can fit this one and the one below in a struct to reduce the number of params here

internal/usecases/liquidity_provider/transfer_excess_to_cold_wallet.go
```diff
@@ -0,0 +1,482 @@
...
+		generalConfig,
+		stateConfig,
+		currentBtcLiquidity,
+		currentRbtcLiquidity,
+	)
+	if err != nil {
+		return nil, usecases.WrapUseCaseError(usecases.TransferExcessToColdWalletId, err)
+	}
+
+	result := &TransferToColdWalletResult{}
```
Luisfc68: prefer the constructors, if you add a new field is easier to update everything if you have a constructor, with this approach you'll need to remember every place where you created the struct

internal/usecases/liquidity_provider/transfer_excess_to_cold_wallet.go
```diff
@@ -0,0 +1,482 @@
...
+
+	return NetworkTransferResult{
+		Status:  TransferStatusFailed,
+		Error:   err,
+		Message: "RSK transfer failed",
+	}
+}
+
+func (useCase *TransferExcessToColdWalletUseCase) validateColdWallet() error {
+	if useCase.coldWallet == nil {
```
Luisfc68: I think this can be assumed as always present, is like having a repository, if you don't have it then the server should not be running, if for some missconfiguration this is not here and the server panics is ok, we don't want it to be running in such conditions

internal/usecases/liquidity_provider/transfer_excess_to_cold_wallet.go
```diff
@@ -0,0 +1,482 @@
...
+		return NoColdWalletConfiguredError
+	}
+
+	btcAddress := useCase.coldWallet.GetBtcAddress()
```
Luisfc68: this one and the one below can be summarized in the same if

internal/usecases/liquidity_provider/transfer_excess_to_cold_wallet.go
```diff
@@ -0,0 +1,482 @@
...
+	if err := generalConfig.ExcessTolerance.Validate(); err != nil {
+		return nil, nil, err
+	}
+
+	stateConfig, err := useCase.getStateConfiguration(ctx)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	if stateConfig.LastBtcToColdWalletTransfer == nil {
```
Luisfc68: I'm not sure about having two last transfer fields... I think is easier and better to have just one, maybe in some iterations it'll pay to only one wallet or both or none, but I don't see the value added in keeping them separated

internal/usecases/liquidity_provider/transfer_excess_to_cold_wallet.go
```diff
@@ -0,0 +1,482 @@
...
+	// AvailablePeginLiquidity returns RBTC balance (from RSK wallet + PegIn contract)
+	rskUsableLiquidity, err := useCase.peginProvider.AvailablePeginLiquidity(ctx)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	return btcUsableLiquidity, rskUsableLiquidity, nil
+}
+
+func (useCase *TransferExcessToColdWalletUseCase) calculateThreshold(
```
Luisfc68: The function seems fine, however, because of the nature of this function I would make it separately (like in the utils/math file) so it can be tested separately, also you're doing percentage/100, to prevent decimal issues, I always recommend scaling the numbers and working with integers, you can check and example in the % fee calculation

internal/usecases/liquidity_provider/transfer_excess_to_cold_wallet.go
```diff
@@ -0,0 +1,482 @@
...
+		forcedExcess := useCase.calculateExcess(target, currentLiquidity, nil)
+		return forcedExcess, true // Transfer due to time forcing
+	}
+
+	return entities.NewWei(0), false
+}
+
+func (useCase *TransferExcessToColdWalletUseCase) shouldForceTransferDueToTime(lastTransferTime *time.Time) bool {
+	secondsSinceLastTransfer := time.Since(*lastTransferTime).Seconds()
+	return secondsSinceLastTransfer >= float64(useCase.forceTransferAfterSeconds)
```
Luisfc68: why to float? Duration comes from int64

internal/usecases/liquidity_provider/transfer_excess_to_cold_wallet.go
```diff
@@ -0,0 +1,482 @@
...
+	excess := useCase.calculateExcess(target, currentLiquidity, threshold)
+
+	if excess.Cmp(entities.NewWei(0)) > 0 {
+		return excess, false // Transfer due to threshold, not time forcing
+	}
+
+	// No excess above threshold, but check if we should force transfer due to time
+	if useCase.shouldForceTransferDueToTime(lastTransferTime) {
+		// Calculate excess without threshold (pass nil)
+		forcedExcess := useCase.calculateExcess(target, currentLiquidity, nil)
```
Luisfc68: This is what I meand with the other comment, IMO calculatedForcedExcess(target, currentLiquidity) >>>>> calculateExcess(target, currentLiquidity, nil)

internal/usecases/liquidity_provider/transfer_excess_to_cold_wallet.go
```diff
@@ -0,0 +1,482 @@
...
+	if amount.Cmp(minWorthwhileAmount) < 0 {
+		return nil, TransferNotEconomicalError
+	}
+
+	txResult, err := useCase.btcWallet.Send(coldBtcAddress, amount)
+	if err != nil {
+		return nil, err
+	}
+
+	return &internalTxResult{
```
Luisfc68: why don't using a generic struct with the structures that we already have for the results in each network (BitcoinTransactionResult and the receipt for rsk)?

internal/usecases/liquidity_provider/transfer_excess_to_cold_wallet.go
```diff
@@ -0,0 +1,482 @@
...
+
+func (useCase *TransferExcessToColdWalletUseCase) executeRskTransfer(ctx context.Context, excess *entities.Wei, isTimeForcingTransfer bool) NetworkTransferResult {
+	if excess.Cmp(entities.NewWei(0)) <= 0 {
+		return NetworkTransferResult{
+			Status:  TransferStatusSkippedNoExcess,
+			Message: "No RSK excess to transfer",
+		}
+	}
+
+	txResult, err := useCase.transferRskExcess(ctx, excess)
```
Luisfc68: same here

internal/adapters/entrypoints/watcher/transfer_cold_wallet_watcher.go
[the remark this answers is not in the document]
AndresQuijano: This feature is not part of this task. I'll create a new task for this.

internal/adapters/entrypoints/rest/routes/routes_test.go
```diff
@@ -255,6 +256,7 @@ func setupRegistryMock(registryMock *mocks.UseCaseRegistryMock) {
 	registryMock.EXPECT().DeleteTrustedAccountUseCase().Return(&liquidity_provider.DeleteTrustedAccountUseCase{})
 	registryMock.EXPECT().RecommendedPegoutUseCase().Return(&pegout.RecommendedPegoutUseCase{})
 	registryMock.EXPECT().RecommendedPeginUseCase().Return(&pegin.RecommendedPeginUseCase{})
+	registryMock.EXPECT().TransferExcessToColdWalletUseCase().Return(&liquidity_provider.TransferExcessToColdWalletUseCase{}).Maybe()
```
Luisfc68: I think you should remove this as no endpoint calls the use case

internal/usecases/liquidity_provider/initialize_state_configuration.go
```diff
@@ -0,0 +1,75 @@
...
+) *InitializeStateConfigurationUseCase {
+	return &InitializeStateConfigurationUseCase{
+		lpRepository: lpRepository,
+		signer:       signer,
+		hashFunc:     hashFunc,
+	}
+}
+
+func (useCase *InitializeStateConfigurationUseCase) Run(ctx context.Context) error {
+	signedStateConfig, err := useCase.lpRepository.GetStateConfiguration(ctx)
```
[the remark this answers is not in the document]
AndresQuijano: My bad. Thanks, I'll change it.

internal/usecases/liquidity_provider/transfer_excess_to_cold_wallet.go
```diff
@@ -0,0 +1,514 @@
...
+	if useCase.shouldForceTransferDueToTime(lastTransferUnix) {
+		// Calculate excess respecting target
+		forcedExcess := useCase.calculateExcess(target, currentLiquidity, target)
+		return forcedExcess, true
+	}
+
+	return entities.NewWei(0), false
+}
+
+func (useCase *TransferExcessToColdWalletUseCase) shouldForceTransferDueToTime(lastTransferUnix *int64) bool {
```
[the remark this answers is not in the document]
AndresQuijano: Also changed in the comment about transfers.

internal/adapters/entrypoints/rest/routes/routes_test.go
AndresQuijano: Yes, removed.

internal/usecases/liquidity_provider/transfer_excess_to_cold_wallet.go
```diff
@@ -0,0 +1,514 @@
...
+	}
+
+	rskResult := useCase.handleRskTransfer(ctx, excessResult.RbtcExcess, excessResult.RbtcIsTimeForced)
+	if rskResult.Status == TransferStatusFailed {
+		result := NewTransferToColdWalletResult(btcResult, rskResult)
+		return result, usecases.WrapUseCaseError(usecases.TransferExcessToColdWalletId, rskResult.Error)
+	}
+
+	result := NewTransferToColdWalletResult(btcResult, rskResult)
+	if btcResult.Status == TransferStatusSuccess || rskResult.Status == TransferStatusSuccess {
```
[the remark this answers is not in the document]
Luisfc68: got it, thanks

internal/usecases/liquidity_provider/transfer_excess_to_cold_wallet.go
```diff
@@ -0,0 +1,546 @@
...
+		BtcExcess:        btcLiquidityExcess,
+		BtcIsTimeForced:  btcIsTimeForced,
+		RbtcExcess:       rbtcLiquidityExcess,
+		RbtcIsTimeForced: rbtcIsTimeForced,
+	}, nil
+}
+
+// publishBtcTransferEvent emits a BTC cold wallet transfer event to the event bus, distinguishing
+// between threshold-triggered and time-forced transfers.
+func (useCase *TransferExcessToColdWalletUseCase) publishBtcTransferEvent(txResult *internalTxResult, isTimeForcingTransfer bool) {
```
Luisfc68: I think you missed `internalTxResult` when removing the pointers from this file (here and in other couple functions)

PR #928, opened 2026-03-06
Feature/FLY-2199 -> Local env modularization

Luisfc68:
## What
Modularization of the local environment

## Why
The current local script is a huge bash script and is hard to maintain. Given the recent frictions with the local env, this PR tries to modularize to extend its flexibility and scalability

### What is in this PR
Reorganization of the directories in the docker compose dir and modularization of the bash script into multiple modules. Check the new README.md added to the local dir to understand how this can be extended

### What is NOT in this PR
- Changes to the general set up logic
- The modules described below

### Posible new modules to add in the short term
- RSK/BTC miners
- Powpeg migration init container
- LBC deployer that pulls from the registry (requires image to be published in the LBC side)
- Flyover UI if we start support custom contract addresses again

## Type of Change
- [ ] Breaking change 
- [ ] Documentation update
- [x] Refactoring (no functional changes, no api changes)
- [ ] Performance improvement
- [ ] Test updates
- [ ] Security fix

## Related Issues
[Jira ticket](https://rsklabs.atlassian.net/browse/FLY-2199)

## How to test
Start the environment using different configurations and enable different combinations of modules (some of the modules are mandatory)


internal/adapters/dataproviders/rootstock/shared_test.go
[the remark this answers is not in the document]
Luisfc68: not part of this PR

docker-compose/local/lps-local.sh
```diff
@@ -0,0 +1,63 @@
+#!/bin/bash
+
```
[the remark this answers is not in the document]
Luisfc68: Added a --reset/-r flag

AndresQuijano (dismissed):

AndresQuijano:
docker-compose/monitoring/grafana/datasources/datasources.yml and docker-compose/monitoring/prometheus/prometheus-config.yml are specific for local env. I think they should be moved to local folder.

docker-compose/local/monitoring/docker-compose.monitoring.yml
AndresQuijano, line 1: A better name for this folder could be monitoring (same as the external)

docker-compose/local/lbc-deployer/scripts/deploy.sh
```diff
@@ -0,0 +1,57 @@
...
+
+if [ -z "$COLLATERAL_PROXY" ] || [ -z "$DISCOVERY_PROXY" ] || [ -z "$PEGIN_PROXY" ] || [ -z "$PEGOUT_PROXY" ]; then
+    echo "ERROR: Failed to parse contract addresses from deployment output"
+    exit 1
+fi
+
+
+echo ""
+echo "Verifying deployed contracts..."
+for CONTRACT_VAR in PEGIN_CONTRACT_ADDRESS PEGOUT_CONTRACT_ADDRESS COLLATERAL_MANAGEMENT_ADDRESS DISCOVERY_ADDRESS; do
```
AndresQuijano: Here you are verifying the old values of the addresses, not the new ones.

docker-compose/local/lbc-deployer/scripts/deploy.sh
```diff
@@ -0,0 +1,57 @@
...
+  CONTRACT_ADDR=$(eval echo "\$$CONTRACT_VAR")
+  CODE=$(curl -s -X POST "$RSK_ENDPOINT" -H "Content-Type: application/json" \
+    -d "{\"jsonrpc\":\"2.0\",\"method\":\"eth_getCode\",\"params\": [\"$CONTRACT_ADDR\",\"latest\"],\"id\":1}" | jq -r ".result")
+
+  if [ "$CODE" = "0x" ] || [ -z "$CODE" ]; then
+    echo "  ✗ $CONTRACT_VAR ($CONTRACT_ADDR) - NO CODE (deployment may have failed)"
```
AndresQuijano, line 38: I think this should exit 1 or something instead of just showing a message

docker-compose/local/lps/configurations/requests.js
```diff
@@ -0,0 +1,90 @@
...
+        }
+    },
+    trustedAccountsConfiguration: {
+        path: '/management/trusted-accounts',
+        method: 'POST',
+        body: {
+            address: process.env.TRUSTED_ACCOUNT_ADDRESS,
+            name: "Boletaz",
+            btcLockingCap: 3000000000000000000,
+            rbtcLockingCap: 3000000000000000000
```
AndresQuijano, line 75: Why these two are not strings?

jormelCoin (approved):
From QA perspective worked well.

docker-compose/local/lps-local.sh
AndresQuijano, line 1: AI review:

The script is missing `set -e`, which was present in the old `lps-env.sh`. This creates three silent failure paths:

**1. Failed healthchecks go undetected**
`docker compose ... up -d --wait` returns non-zero if any container fails to become healthy (e.g. LPS crashes on startup, bitcoind never starts). Without `set -e` the script continues and prints `✓ LPS environment is ready!` even when it isn't.

**2. `docker wait` exit code is discarded (line 51)**
```bash
docker wait lbc-deployer
echo "Contracts deployed"
```
`docker wait` prints the container's exit code to stdout but doesn't fail the shell. If `deploy.sh` exits with `1` the script still prints "Contracts deployed", sources the unchanged env file, and starts LPS with stale contract addresses. This should be:
```bash
EXIT_CODE=$(docker wait lbc-deployer)
if [ "$EXIT_CODE" != "0" ]; then
  echo "ERROR: Contract deployment failed (exit code $EXIT_CODE)"
  exit 1
fi
```

**3. Wallet funder failure is silent**
If `funder.js` throws, the compose on line 44 returns non-zero — ignored without `set -e`. Contract deployment then runs against unfunded wallets, fails on-chain, and that failure is also ignored (see point 2).

Suggested fix: add `set -e` after the env file is sourced (line 36), so the reset block is unaffected, and add the explicit exit code check for `docker wait`.

README.md
```diff
@@ -22,8 +22,7 @@ To run the project locally you can follow these steps:
 
 1. `git clone git@github.com:rsksmart/liquidity-provider-server.git`
 2. `cd docker-compose/local`
-3. `export LPS_STAGE=regtest`
-4. `./lps-env.sh up`
+3`./lps-local.sh up`
```
AndresQuijano: The script doesn't accept the "up" argument

AndresQuijano: There is also a small typo on that line. Instead of 3, it should be 3. so the format is the same as the other items of the numbered list

docker-compose/local/lps/configurations/requests.js
Luisfc68: because the type in the DTO is big.Int, not string, maybe it supports both I'd need to check, but I rather use this representation as its of the same type, additionally emulates the sdk behavior since the bigint serializer we use there serializes in the same way (without the quotes)

docker-compose/local/docker-compose.yml
```diff
@@ -1,8 +1,7 @@
-version: "3"
 services:
   localstack:
     build:
-      context: ../localstack
+      context: ./localstack
       args:
         UID: "${LPS_UID}"
```
AndresQuijano, line 6: After this line I propose to do pull_policy: build for localstack, bitcoind and rskj so it does not try to pull the image and goes directly to build. The file as I propose would start like this:
`services:
  localstack:
    build:
      context: ./localstack
      args:
        UID: "${LPS_UID}"
    pull_policy: build
    container_name: localstack
    image: flyover-localstack
    ports:
      - "127.0.0.1:4566:4566"            # LocalStack Gateway
      - "127.0.0.1:4510-4559:4510-4559"  # external services port range
    environment:
      - DEBUG=${DEBUG:-0}
    volumes:
      - "${LOCALSTACK_HOME:-./volumes/localstack}:/var/lib/localstack"
      - "/var/run/docker.sock:/var/run/docker.sock"
    networks:
      - net_lps
  bitcoind:
    build:
      context: ../bitcoind
      args:
        UID: "${LPS_UID}"
    pull_policy: build
    image: bitcond:latest
    container_name: bitcoind01
    environment:
      - BTC_USERNAME
      - BTC_PASSWORD
    ports:
      - "5555:5555"
    volumes:
      - ${BTCD_HOME:-./volumes/bitcoind}:/home/bitcoind/.bitcoind
    networks:
      - net_lps
    command:
      [
        "bitcoind",
        "-rpcuser=${BTC_USERNAME}",
        "-rpcpassword=${BTC_PASSWORD}",
        "-addresstype=legacy",
        "-${LPS_STAGE}",
        "-printtoconsole",
        "-server",
        "-txindex",
        "-deprecatedrpc=signrawtransaction",
        "-deprecatedrpc=accounts",
        "-rpcbind=0.0.0.0",
        "-rpcallowip=0.0.0.0/0",
        "-rpcport=5555",
      ]
  rskj:
    build:
      context: ../rskj
      args:
        UID: "${LPS_UID}"
    pull_policy: build
    image: rskj:latest
    container_name: rskj01
    ports:
      - "4444:4444"
      - "4445:4445"
    volumes:
      - ${RSKJ_HOME:-./volumes/rskj}/db:/var/lib/rsk/database
      - ${RSKJ_HOME:-./volumes/rskj}/logs:/var/log/rsk
    networks:
      - net_lps
    command:
      [
        "--${LPS_STAGE}",
      ]`
....rest of the file

docker-compose/local/lbc-deployer/docker-compose.lbc-deployer.yml
```diff
@@ -0,0 +1,25 @@
+services:
+  lbc-deployer:
+    build:
+      context: ./lbc-deployer
+      args:
+        LBC_COMMIT: "${LBC_COMMIT}"
+    image: lbc-deployer:latest
```
AndresQuijano, line 7: Same after this line:
pull_policy: build

AndresQuijano: The docker-compose/local/lps/docker-compose.lps-local.yml was not modified in this PR but I also recommend doing the same thing.
Same for docker-compose/local/lbc-deployer/docker-compose.lbc-deployer.yml, and docker-compose/local/powpeg/docker-compose.powpeg.yml
I want to say that this is just a recommendation because it eliminates a lot of noise in the output of the script

AndresQuijano (approved):

PR #934, opened 2026-03-13
Add configurable rebalance strategy for bridge pegout

volodymyrzahanych:
## What
Add configurable rebalance strategy for bridge pegout, supporting two modes: ALL_AT_ONCE (existing behavior - converts entire RSK balance to BTC in a single transaction) and UTXO_SPLIT (splits the pegout amount into multiple UTXOs based on the provider's min/max pegout limits for better liquidity distribution)

## Why
The current rebalance behavior sends all funds waiting for rebalance in a single native pegout transaction, causing UTXOs to converge into one during usage spikes
When all liquidity is consumed, only one large native pegout is executed, reducing UTXO availability for subsequent operations

## Type of Change
- [ ] Breaking change 
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test updates
- [ ] Security fix

- [ ] Management UI / API
- [ ] PegIn flow
- [x] PegOut flow
- [ ] Utility scripts
- [ ] Configuration files
- [ ] Metrics and alerting

## Related Issues
[Jira ticket](https://rsklabs.atlassian.net/browse/FLY-2235)

## How to test
Set REBALANCE_STRATEGY=UTXO_SPLIT, initiate a large pegout (e.g. 3 RBTC), wait for rebalance, and verify the LP BTC wallet receives multiple UTXOs of ~BridgeTransactionMin size instead of one large UTXO
Set REBALANCE_STRATEGY=ALL_AT_ONCE, repeat the same flow, and verify the LP BTC wallet receives a single UTXO with the full amount
Run unit tests: go test ./internal/usecases/pegout/ -run TestBridgePegout

## Reviewer Guidelines
- [ ] Verify the splitting math: `firstChunk + (N-1)×bridgeMin = totalValue`
- [ ] Review partial failure handling — acceptable trade-off?
- [ ] Check if `updateQuotes` storing only last tx hash  needs a follow-up ticket
- [ ] Confirm `BridgeTransactionMin` in management API matches the Bridge's actual minimum
- [ ] Verify the 66-tx scenario (100 RBTC test case) is acceptable for production Bridge load
- [ ] Decide if discarded error from `Div()` should use explicit `_` or add a comment
- [ ] Test locally with `REBALANCE_STRATEGY=UTXO_SPLIT` in `.env.regtest` and run `pegout-cycle.sh full-cycle --repeat 3` + `check-btc-release`

### AI Code Review: Configurable Rebalance Strategy for Bridge Pegout

**Commit:** `5cea8c2` — "Add configurable rebalance strategy for bridge pegout"
**Branch:** `feature/FLY-2235` 
**Reviewer:** Claude Opus 4.6 (AI-assisted review)
**Date:** 2026-03-13

---

#### 1. Summary

This feature introduces a configurable rebalance strategy for the bridge pegout flow. Previously, the LP always sent the entire accumulated refund total to the RSK Bridge in a single transaction (`ALL_AT_ONCE`). The new `UTXO_SPLIT` strategy splits the bridge conversion into multiple transactions of `BridgeTransactionMin` size each, producing multiple UTXOs on the BTC side.

**Motivation:** When the Bridge releases BTC back to the LP, each bridge transaction produces one UTXO. With `UTXO_SPLIT`, the LP receives N smaller UTXOs instead of one large one, improving UTXO management for future pegout operations.

---

#### 2. Files Changed

| File | Change |
|------|--------|
| `internal/usecases/pegout/bridge_pegout.go` | Core logic: `RebalanceStrategy` type, `ParseRebalanceStrategy()`, `runAllAtOnce()`, `runUtxoSplit()` |
| `internal/usecases/pegout/bridge_pegout_test.go` | 6 UTXO_SPLIT tests + 8-case amount integrity test |
| `internal/configuration/environment/environment.go` | Added `RebalanceStrategy` field to `PegoutEnv` |
| `internal/configuration/environment/environment_reader_test.go` | Test env setup |
| `internal/configuration/registry/usecase.go` | Parse strategy at startup, pass to `BridgePegoutUseCase` |
| `internal/configuration/registry/usecase_test.go` | Test setup |
| `internal/configuration/registry/watcher_test.go` | Test setup |
| `internal/adapters/entrypoints/watcher/pegout_bridge_watcher_test.go` | Updated constructor call |
| `cmd/application/lps/application.go` | Handle `NewUseCaseRegistry` returning `(*UseCaseRegistry, error)` |
| `docker-compose/local/docker-compose.lps.yml` | Added `REBALANCE_STRATEGY` to container env |
| `sample-config.env` | Default: `REBALANCE_STRATEGY=ALL_AT_ONCE` |

---

#### 3. Architecture & Data Flow

```
Environment                    Registry                       Use Case
─────────────────────────────────────────────────────────────────────────
REBALANCE_STRATEGY env var
        │
        ▼
PegoutEnv.RebalanceStrategy
  (validated: required,
   oneof=ALL_AT_ONCE
         UTXO_SPLIT)
        │
        ▼
                         ParseRebalanceStrategy()
                               │
                               ▼
                         NewBridgePegoutUseCase(..., strategy)
                               │
                               ▼
                         BridgePegoutUseCase.strategy (immutable field)
                               │
                               ▼
              ┌────────────────┴────────────────┐
              │                                 │
        ALL_AT_ONCE                       UTXO_SPLIT
              │                                 │
              ▼                                 ▼
     runAllAtOnce()                    runUtxoSplit()
     1 × SendRbtc(total)              N × SendRbtc(bridgeMin)
                                       (first chunk = bridgeMin + remainder)
```

### Trigger Path

1. `PegoutBridgeWatcher` ticks every 5 minutes
2. Queries quotes in `RefundPegOutSucceeded` state
3. Calls `BridgePegoutUseCase.Run(ctx, quotes...)`
4. `Run()` calculates `totalValue = Σ(Value + CallFee + GasFee)` for all quotes
5. If `totalValue < BridgeTransactionMin` → skip (normal, logged as info)
6. Acquires `rskWalletMutex`
7. Delegates to `runAllAtOnce()` or `runUtxoSplit()` based on `strategy`

---

## 4. Splitting Algorithm (UTXO_SPLIT)

```
Input:  totalValue, bridgeMin (from PegoutConfiguration)

numTxs    = totalValue / bridgeMin          (integer division, truncated)
remainder = totalValue - (numTxs × bridgeMin)

Invariant: numTxs × bridgeMin + remainder = totalValue (exact, no rounding)

If N = 1:
    Send 1 tx of totalValue (no split needed)

If N ≥ 2:
    Tx 1:     bridgeMin + remainder    (first chunk absorbs the remainder)
    Tx 2..N:  bridgeMin each

Sum = (bridgeMin + remainder) + (N-1) × bridgeMin
    = N × bridgeMin + remainder
    = totalValue                       ✓ exact
```

**Why this is rounding-safe:** All arithmetic uses Go's `math/big.Int` (arbitrary precision integers). The remainder is computed by subtraction (`total - N×min`) rather than modulo, so `firstChunk + (N-1)×min = total` is an algebraic identity, not an approximation.

### Gas Calculation

```
gasPerTx        = BridgeConversionGasLimit × BridgeConversionGasPrice = 100000 × 60000000 = 6×10¹² wei
requiredBalance = totalValue + N × gasPerTx
```

For `ALL_AT_ONCE`, gas is `1 × gasPerTx`. For `UTXO_SPLIT` with N transactions, gas scales linearly as `N × gasPerTx`.

---

### 5.1 Correctness

| # | Item | Status | Notes |
|---|------|--------|-------|
| 1 | Integer division produces correct N | ✅ | `big.Int.Div` truncates toward zero |
| 2 | Remainder computation is exact | ✅ | `remainder = total - N×min` (subtraction, not modulo) |
| 3 | Sum of all sent amounts = totalValue | ✅ | Algebraic identity, verified by `TestUtxoSplit_AmountIntegrity` |
| 4 | Each split tx ≥ bridgeMin | ✅ | First chunk = `min + remainder ≥ min`; rest = `min` |
| 5 | N=1 sends totalValue (not bridgeMin) | ✅ | Line 148: `config := blockchain.NewTransactionConfig(totalValue, ...)` |
| 6 | Balance check accounts for N×gas | ✅ | Line 137: `requiredBalance = totalValue + N×gasPerTx` |
| 7 | `bridgeMin.Copy()` prevents mutation | ✅ | Lines 161, 178 use `.Copy()` before passing to tx config |
| 8 | Strategy dispatch uses switch/default | ✅ | `ALL_AT_ONCE` is the default (line 94) |

### 5.2 Error Handling

| # | Scenario | Behavior | Assessment |
|---|----------|----------|------------|
| 1 | Quotes not all refunded | Error returned early, no bridge tx | ✅ correct |
| 2 | Total below bridgeMin | `TxBelowMinimumError`, logged as info by watcher | ✅ correct |
| 3 | Wallet balance error | Error returned, quotes not updated | ✅ correct |
| 4 | Insufficient balance | `InsufficientAmountError` returned | ✅ correct |
| 5 | First tx fails (N≥2) | Quotes marked `BridgeTxFailed`, error returned | ✅ correct |
| 6 | Tx K fails mid-split (K>1) | Loop breaks, quotes marked failed using last receipt | ⚠️ see 5.3 |
| 7 | `UpdateRetainedQuotes` fails | Error returned (joined with tx error if any) | ✅ correct |

#### A. Partial Success Handling in UTXO_SPLIT

When N≥2 and tx K fails (K>1), transactions 1..K-1 have already been sent to the bridge and **cannot be rolled back**. The code marks all quotes as `BridgeTxFailed`:

```go
// bridge_pegout.go:177-185
for i := uint64(1); i < n; i++ {
    config = blockchain.NewTransactionConfig(bridgeMin.Copy(), ...)
    receipt, txErr = useCase.rskWallet.SendRbtc(ctx, config, bridgeAddress)
    if txErr == nil {
        log.Debugf(...)
    } else {
        break
    }
}
err = useCase.updateQuotes(ctx, receipt, txErr, watchedQuotes)
```

**What happens:**
- RBTC from successful txs has already been sent to the Bridge — those funds are in transit and will eventually produce BTC UTXOs
- But all quotes are marked `BridgeTxFailed`, so the watcher won't re-process them
- The Bridge will still release BTC for the successful txs (bridge is stateless w.r.t. LPS quote tracking)

**Assessment:** This is a **conservative design choice** — it prevents double-sending by marking all quotes as terminal. The LP receives the partial BTC but the accounting is pessimistic. An operator would need to reconcile manually.

**Alternative (not necessarily better):** Track per-tx success and mark quotes proportionally. But this adds complexity and partial state management that may not be worth it given the rarity of mid-split failures.

#### B. `updateQuotes` Uses Last Receipt Only

In the split loop, `updateQuotes` is called once with the final receipt:

```go
// bridge_pegout.go:187
err = useCase.updateQuotes(ctx, receipt, txErr, watchedQuotes)
```

On success (all N txs complete), `receipt` is from the last tx. This means `BridgeRefundTxHash` on all quotes points to tx N, not tx 1. For `ALL_AT_ONCE` this is fine (one tx). For `UTXO_SPLIT`, it means quote records reference only the last bridge tx hash.

**Impact:** The other tx hashes are logged (debug level) but not persisted in the database. If an operator needs to audit all bridge transactions for a rebalance cycle, they'd need to check logs or chain state.

#### C. `numTxs` Returned as `*Wei` From Division

```go
numTxs, _ := new(entities.Wei).Div(totalValue, bridgeMin)
```

The error from `Div` is discarded. This is safe because `bridgeMin` is validated upstream (it passes the `BridgeTransactionMin.Cmp(totalValue) > 0` check at line 79, confirming `bridgeMin > 0`), so division by zero cannot occur. However, the discarded error is a minor code smell.

#### D. Default Strategy is ALL_AT_ONCE

The `switch` statement defaults to `runAllAtOnce`:

```go
switch useCase.strategy {
case UtxoSplit:
    return useCase.runUtxoSplit(...)
default:
    return useCase.runAllAtOnce(...)
}
```

This is safe: even if a new strategy value were somehow injected, the system falls back to the original single-tx behavior. The double validation (env + parsing) prevents this in practice.

---

### Existing Tests (from the cherry-picked commit)

| Test | Strategy | Scenario | Verifies |
|------|----------|----------|----------|
| `testBridgePegoutUseCaseSuccess` | ALL_AT_ONCE | Happy path, total=558 | Exact amount, receipt fields, state transition |
| `testBridgePegoutUseCaseValueBelowMinimum` | ALL_AT_ONCE | total < min | No wallet calls, correct error |
| `testBridgePegoutUseCaseQuotesNotRefunded` | ALL_AT_ONCE | Mixed quote states | Early rejection |
| `testBridgePegoutUseCaseWalletBalanceError` | ALL_AT_ONCE | GetBalance error | Error propagated |
| `testBridgePegoutUseCaseWalletWithoutBalance` | ALL_AT_ONCE | balance < required | InsufficientAmountError |
| `testBridgePegoutUseCaseTxFails` | ALL_AT_ONCE | SendRbtc error | Quotes marked BridgeTxFailed |
| `testBridgePegoutUseCaseUpdateFails` | ALL_AT_ONCE | DB update error | Error returned |
| `testUtxoSplitSuccess` | UTXO_SPLIT | N=2, R=158 | Tx amounts: 358+200 |
| `testUtxoSplitNoSplitWhenN1` | UTXO_SPLIT | N=1 | Single tx of totalValue |
| `testUtxoSplitBelowMinimum` | UTXO_SPLIT | total < min | Rejected |
| `testUtxoSplitExactMultiple` | UTXO_SPLIT | N=3, R=0 | Three equal txs of 200 |
| `testUtxoSplitFailMidSplit` | UTXO_SPLIT | N=2, tx2 fails | All quotes marked failed |
| `testUtxoSplitInsufficientGas` | UTXO_SPLIT | N=2, gas too low | InsufficientAmountError |
| `TestParseRebalanceStrategy` | — | Parsing | Valid/invalid strategy strings |

### Added Tests (on this branch)

| Test | Scenario | Key Assertion |
|------|----------|---------------|
| `TestUtxoSplit_AmountIntegrity` (8 sub-tests) | Various N and remainders at wei scale | **sum(all sent amounts) == totalValue exactly** + each amount ≥ bridgeMin |

Cases in the integrity test:

| Sub-test | Total | Min | N | Key Verification |
|----------|-------|-----|---|------------------|
| total == min | 1.5 RBTC | 1.5 RBTC | 1 | Single tx boundary |
| just above min | 2.9 RBTC | 1.5 RBTC | 1 | N=1 absorbs all |
| exact 2x | 3.0 RBTC | 1.5 RBTC | 2 | Zero remainder, exact split |
| just below 2x | 2.9 RBTC | 1.5 RBTC | 1 | Integer truncation keeps N=1 |
| 2x + 1 wei | 3.0 RBTC + 1 wei | 1.5 RBTC | 2 | Smallest possible remainder |
| large remainder | 8.3 RBTC | 1.5 RBTC | 5 | 0.8 RBTC remainder in first chunk |
| exact 3x | 4.5 RBTC | 1.5 RBTC | 3 | Multi-tx zero remainder |
| 100 RBTC / 66 txs | 100 RBTC | 1.5 RBTC | 66 | Large N with 1.0 RBTC remainder |

### Coverage Gaps (Minor)

1. **No test with multiple quotes and UTXO_SPLIT** — all UTXO_SPLIT tests use a single quote or the shared `bridgePegoutTestWatchedQuotes`. A test with many quotes where Value+CallFee+GasFee produces an interesting total would add confidence.
2. **No test for first tx failure in N≥2** — `testUtxoSplitFailMidSplit` tests tx2 failure. A test where tx1 itself fails would verify the early-exit path at lines 170-175.

---

### 7.1 LPS Flow Security

| Concern | Assessment | Risk |
|---------|------------|------|
| **Fund safety** | All bridge transactions target `contracts.Bridge.GetAddress()` — hardcoded from contract registry, not user-controlled | ✅ None |
| **Double-send prevention** | `rskWalletMutex` serializes all bridge sends; quotes transition to terminal state (`BridgeTxSucceeded`/`BridgeTxFailed`) atomically | ✅ None |
| **Denial of service via config** | `REBALANCE_STRATEGY` validated at startup — invalid value prevents boot. No runtime mutation | ✅ None |
| **Integer overflow** | `big.Int` arithmetic — no overflow possible | ✅ None |
| **Rounding/loss of funds** | Sum invariant proven algebraically and verified by tests with wei-precision values | ✅ None |
| **Gas exhaustion** | Balance check includes `N × gasPerTx` before any send. If insufficient, rejects early | ✅ None |
| **Partial failure state** | Conservative: all quotes marked failed on any mid-split error. No funds are lost — successful bridge txs still produce BTC releases | ⚠️ Low (accounting mismatch, not fund loss) |
| **Strategy injection** | Double-validated (env validator + switch/case). Default falls back to `ALL_AT_ONCE`. No path to inject arbitrary strategy | ✅ None |

### 7.2 Rootstock Federation Bridge Flow Security

This section examines how the UTXO_SPLIT strategy affects the RSK Bridge and federation nodes.

#### Bridge Perspective

The RSK Bridge at `0x0000000000000000000000000000000001000006` processes incoming RBTC transfers as pegout requests. Each transfer that meets the minimum lock value is queued for BTC release.

| Concern | Analysis |
|---------|----------|
| **Multiple small txs vs one large tx** | The Bridge treats each incoming transfer independently. Whether the LP sends 1×4.5 RBTC or 3×1.5 RBTC, the Bridge queues them separately. This is standard Bridge behavior — no new attack surface. |
| **Bridge minimum enforcement** | Each UTXO_SPLIT transaction sends exactly `bridgeMin` (or `bridgeMin + remainder` for the first). Since `bridgeMin` is configured to match `BridgeTransactionMin` (the Bridge's own minimum), all txs meet the Bridge's acceptance threshold. |
| **Pegout queue congestion** | UTXO_SPLIT creates N pegout queue entries instead of 1. On mainnet, the Bridge batches pegouts per block interval (`getNextPegoutCreationBlockNumber`). More queue entries mean more BTC outputs in the release tx, increasing its size. For N≤10 this is negligible; for very large N (e.g., 66 in the 100 RBTC test case), the Bridge release tx could be significant. |
| **Federation signing load** | Each queued pegout requires federation consensus for the BTC release. N pegouts in the same batch are signed together (single BTC tx with N outputs), so the signing overhead is per-batch, not per-pegout. No significant impact. |

#### BTC Side

| Concern | Analysis |
|---------|----------|
| **UTXO set for the LP** | This is the intended benefit — N UTXOs of ~`bridgeMin` BTC each, instead of 1 large UTXO. Improves the LP's ability to serve future pegout requests without change-output management. |
| **BTC transaction fee for Bridge release** | A Bridge release tx with N outputs costs more in BTC fees than one with 1 output. The fee comes from the Bridge's fee-per-KB calculation (`getFeePerKb`). For typical N (2-5), this is minimal. The LP does not pay this fee — it's deducted from the Bridge's pool. |
| **Dust outputs** | Not possible: each output is at least `bridgeMin` (1.5 RBTC = 1.5 BTC in regtest), far above Bitcoin's dust threshold (~546 satoshi). |

#### Timing & Ordering

| Concern | Analysis |
|---------|----------|
| **Nonce sequencing** | UTXO_SPLIT sends N transactions sequentially from the same RSK wallet. The `rskWalletMutex` ensures serialization. Each `SendRbtc` call should increment the nonce atomically. If the wallet implementation handles nonces correctly, this is safe. |
| **Block inclusion** | N txs may land in the same RSK block or span multiple blocks. The Bridge processes them by block, so they may appear in the same or different pegout batches. This is acceptable — the Bridge handles this natively. |
| **Race with other pegout cycles** | The watcher ticks every 5 minutes. The `rskWalletMutex` prevents concurrent bridge sends. A split of 66 txs at the gas price/limit used would complete well within 5 minutes. No race condition. |

### 7.3 Attack Scenarios Considered

| Attack | Feasibility | Mitigation |
|--------|-------------|------------|
| **Attacker sets REBALANCE_STRATEGY to malicious value** | Requires access to the LPS container environment. Env validation rejects unknown values at startup. | Double validation + startup failure |
| **Attacker manipulates BridgeTransactionMin to force many small txs** | BridgeTransactionMin is set via authenticated Management API with CSRF protection. Not externally accessible. | Authentication + CSRF |
| **Attacker triggers excessive bridge congestion via UTXO_SPLIT** | LP operator controls their own strategy. An attacker cannot force the LP to use UTXO_SPLIT. The Bridge handles multiple pegout entries natively. | Operator-only config |
| **Front-running bridge txs during split** | Attacker could insert a tx between split txs. But the Bridge processes all qualifying transfers regardless of ordering. No exploit possible. | Bridge is order-independent |
| **Grief the LP with partial failure** | If the RSK network rejects mid-split txs (e.g., gas price spike), the LP marks all quotes as failed. LP funds from successful txs are not lost — they produce BTC UTXOs. The LP needs manual reconciliation. | Conservative failure handling |

### 7.4 Security Verdict

**No security vulnerabilities identified.** The feature is a clean extension of the existing rebalance flow. The `UTXO_SPLIT` strategy interacts with the Bridge in the same way as `ALL_AT_ONCE` — it simply sends multiple qualifying transfers instead of one. The Bridge's native handling of pegout queues, fee calculation, and federation signing is unaffected.

The one operational risk (partial failure accounting mismatch) is a design trade-off, not a vulnerability. It favors safety (no double-send) over convenience (manual reconciliation).

---

## 8. Compatibility with This Branch

The main feature commit was tested on `master` and then moved onto `feature/FLY-2235`. Key differences on this branch:

| Branch Difference | Impact on Rebalance Feature |
|-------------------|-----------------------------|
| `LBC_ADDR` replaced by split contract addresses | None — rebalance uses `contracts.Bridge`, not LBC |
| `FeeCollectorAddress` removed | None — rebalance does not reference fee collector |
| `ProviderType` changed from field to method | None — rebalance does not reference provider type |
| `AllowedOrigins` CORS field added | None — unrelated to use case layer |
| `NewUseCaseRegistry` return type change | Correctly applied — returns `(*UseCaseRegistry, error)` |

**All unit tests pass** on the branch (32 packages). Build compiles cleanly. No merge conflicts.


internal/configuration/environment/environment.go
```diff
@@ -168,6 +168,7 @@ type PegoutEnv struct {
 	DepositCacheStartBlock      uint64 `env:"PEGOUT_DEPOSIT_CACHE_START_BLOCK"`
 	BtcReleaseWatcherStartBlock uint64 `env:"BTC_RELEASE_WATCHER_START_BLOCK"`
 	BtcReleaseWatcherPageSize   uint64 `env:"BTC_RELEASE_WATCHER_PAGE_SIZE"`
+	RebalanceStrategy           string `env:"REBALANCE_STRATEGY" validate:"omitempty,oneof=ALL_AT_ONCE UTXO_SPLIT"`
```
Luisfc68: I think this config should be mandatory, no need of allowing an empty value, because we don't have a criteria to pick a default strategy

```suggestion
	RebalanceStrategy           string `env:"REBALANCE_STRATEGY" validate:"oneof=ALL_AT_ONCE UTXO_SPLIT"`
```

docker-compose/local/docker-compose.lps.yml
```diff
@@ -87,6 +87,7 @@ services:
       - BTC_RELEASE_WATCHER_START_BLOCK
       - BTC_RELEASE_WATCHER_PAGE_SIZE
       - BTC_RELEASE_CHECK_TIMEOUT
+      - REBALANCE_STRATEGY=${REBALANCE_STRATEGY:-ALL_AT_ONCE}
```
Luisfc68: In my other comment I suggested to enforce this variable to be provided, but if you still prefer to have a default value my advice is to define it inside the code, as we don't have a guarantee that the server will be deployed using the compose files

internal/usecases/pegout/bridge_pegout.go
```diff
@@ -20,12 +21,31 @@ const (
...
+type RebalanceStrategy string
+
+const (
+	AllAtOnce RebalanceStrategy = "ALL_AT_ONCE"
+	UtxoSplit RebalanceStrategy = "UTXO_SPLIT"
+)
+
+func ParseRebalanceStrategy(s string) (RebalanceStrategy, error) {
+	switch s {
+	case "", string(AllAtOnce):
```
Luisfc68: I think adding the empty string here makes the function less predictable, if you want to add the default behavior it can be in a layer above but in this case you're just trying to get the `RebalanceStrategy` from a string

internal/usecases/pegout/bridge_pegout.go
```diff
@@ -66,23 +88,94 @@ func (useCase *BridgePegoutUseCase) Run(ctx context.Context, watchedQuotes ...qu
 	useCase.rskWalletMutex.Lock()
 	defer useCase.rskWalletMutex.Unlock()
 
-	requiredBalance := new(entities.Wei).Add(totalValue, entities.NewWei(BridgeConversionGasLimit*BridgeConversionGasPrice))
-	if balance, err = useCase.rskWallet.GetBalance(ctx); err != nil {
+	switch useCase.strategy {
```
Luisfc68: I think this approach won't scale properly, every time we need to add a new strategy we'll need to add one switch statement here, one in the function that parses the string, and the functionality in this use case. I think the proper way to implement this is through an interface, a factory and the factory would be the one in charge of routing the behavior to the implementation. Additionally, with that approach we can have more complex rebalancing strategies (as its very likely to happen) because the rebalancing strategy implementation would be a struct itself, so it could have state, dependencies, etc

internal/usecases/pegout/bridge_pegout.go
```diff
@@ -66,23 +88,94 @@ func (useCase *BridgePegoutUseCase) Run(ctx context.Context, watchedQuotes ...qu
...
+		config = blockchain.NewTransactionConfig(bridgeMin.Copy(), BridgeConversionGasLimit, entities.NewWei(BridgeConversionGasPrice))
+		receipt, txErr = useCase.rskWallet.SendRbtc(ctx, config, bridgeAddress)
+		if txErr == nil {
+			log.Debugf("%s: split tx %d/%d sent to the bridge successfully (%s)", usecases.BridgePegoutId, i+1, numTxs, receipt.TransactionHash)
+		} else {
+			break
+		}
+	}
+
+	if err := useCase.updateQuotes(ctx, receipt, txErr, watchedQuotes); err != nil {
```
Luisfc68: there is a problem here which is that all the quotes will use the same receipt for the update, each quote should have the information of its own rebalance tx in the database

Luisfc68: I mean, you're making N transactions but only passing one receipt

internal/usecases/pegout/bridge_pegout.go
```diff
@@ -66,23 +88,94 @@ func (useCase *BridgePegoutUseCase) Run(ctx context.Context, watchedQuotes ...qu
...
+	var txErr error
+
+	config := blockchain.NewTransactionConfig(firstChunk, BridgeConversionGasLimit, entities.NewWei(BridgeConversionGasPrice))
+	receipt, txErr = useCase.rskWallet.SendRbtc(ctx, config, bridgeAddress)
+	if txErr == nil {
+		log.Debugf("%s: split tx 1/%d sent to the bridge successfully (%s)", usecases.BridgePegoutId, numTxs, receipt.TransactionHash)
+	} else {
+		return usecases.WrapUseCaseError(
+			usecases.BridgePegoutId,
+			useCase.updateQuotes(ctx, receipt, txErr, watchedQuotes),
```
Luisfc68: same here you're updating all the quotes with only 1 receipt

internal/configuration/registry/usecase.go
```diff
@@ -92,7 +92,11 @@ func NewUseCaseRegistry(
 	liquidityProvider *dataproviders.LocalLiquidityProvider,
 	messaging *Messaging,
 	mutexes entities.ApplicationMutexes,
-) *UseCaseRegistry {
+) (*UseCaseRegistry, error) {
+	rebalanceStrategy, err := pegout.ParseRebalanceStrategy(env.Pegout.RebalanceStrategy)
```
Luisfc68: this is what I mean in my comment in the other file, if you implement this using interfaces and a factory, you could call the factory here and then the use case would receive an usable struct, so it wouldn't need to switch there

Luisfc68 (changes requested):

internal/configuration/environment/environment.go
volodymyrzahanych: Actually it was my first idea to make But then copilot said in its review:

> Making REBALANCE_STRATEGY required will cause the application to fail startup whenever the variable isn’t set (empty string), even though the intended default is the existing behavior. To preserve backward compatibility and simplify deployments, consider providing a default (e.g., treat empty as ALL_AT_ONCE) and/or removing the required validator while still validating allowed values when set.

I revert to the required version. It is just simplier .

volodymyrzahanych: Actually it was my first idea to make it required. But then copilot said in its review:

> Making REBALANCE_STRATEGY required will cause the application to fail startup whenever the variable isn’t set (empty string), even though the intended default is the existing behavior. To preserve backward compatibility and simplify deployments, consider providing a default (e.g., treat empty as ALL_AT_ONCE) and/or removing the required validator while still validating allowed values when set.

I revert to the required version. It is just simplier .

docker-compose/local/docker-compose.lps.yml
volodymyrzahanych: Done — removed the default value so the env var is just passed through, consistent with it being required.

Luisfc68:
@volodymyrzahanych one comment I forgot: there are multiple compose files for the LPS, pls add the env var to all of them, right now is just in the local. Thanks!

internal/usecases/pegout/bridge_pegout.go
volodymyrzahanych: Done — removed the empty string case from ParseRebalanceStrategy.

internal/entities/quote/pegout_quote.go
```diff
@@ -127,23 +127,31 @@ func (quote *PegoutQuote) Total() *entities.Wei {
...
+	Signature            string                      `json:"signature" bson:"signature" validate:"required"`
+	RequiredLiquidity    *entities.Wei               `json:"requiredLiquidity" bson:"required_liquidity" validate:"required"`
+	State                PegoutState                 `json:"state" bson:"state" validate:"required"`
+	UserRskTxHash        string                      `json:"userRskTxHash" bson:"user_rsk_tx_hash"`
+	LpBtcTxHash          string                      `json:"lpBtcTxHash" bson:"lp_btc_tx_hash"`
+	RefundPegoutTxHash   string                      `json:"refundPegoutTxHash" bson:"refund_pegout_tx_hash"`
+	BridgeRefundTxHash   string                      `json:"BridgeRefundTxHash" bson:"bridge_refund_tx_hash"`
+	BridgeRefundGasUsed  uint64                      `json:"bridgeRefundGasUsed" bson:"bridge_refund_gas_used"`
+	BridgeRefundGasPrice *entities.Wei               `json:"bridgeRefundGasPrice" bson:"bridge_refund_gas_price"`
+	RemainingToRefund    *entities.Wei               `json:"remainingToRefund,omitempty" bson:"remaining_to_refund,omitempty"`
```
Luisfc68: I think the usage of the RemainingToRefund field is not obvious, it would be nice if you add a comment to document its usage. Additionally, I recommend to update the creation of the RetainedPegoutQuote to set this by default to the same value of `quote.Total()`

internal/entities/quote/pegout_quote.go
```diff
@@ -127,23 +127,31 @@ func (quote *PegoutQuote) Total() *entities.Wei {
...
+	BridgeRebalances     []BridgeRebalanceAllocation `json:"bridgeRebalances,omitempty" bson:"bridge_rebalances,omitempty"`
```
Luisfc68: I rather avoid using `omitempty` without a specific reason, specially in this field (and any other that uses pointers), because I don't know if when this struct is read from the db it will be read as nil instead of empty slice, which can increase the chances of nil pointer errors accross the application

internal/configuration/registry/usecase.go
```diff
@@ -92,7 +92,7 @@ func NewUseCaseRegistry(
 	liquidityProvider *dataproviders.LocalLiquidityProvider,
 	messaging *Messaging,
 	mutexes entities.ApplicationMutexes,
-) *UseCaseRegistry {
+) (*UseCaseRegistry, error) {
```
Luisfc68: if there is no chance this returns an error anymore you should update the signature

internal/usecases/pegout/bridge_pegout.go
```diff
@@ -20,115 +18,91 @@ const (
...
-	if pegoutConfig.BridgeTransactionMin.Cmp(totalValue) > 0 {
-		log.Infof(
-			"Refunded pegouts total value: %v out of %v. Threshold not met yet. Skipping transaction to the bridge.",
-			totalValue.ToRbtc(),
-			pegoutConfig.BridgeTransactionMin.ToRbtc(),
-		)
-		return usecases.WrapUseCaseError(usecases.BridgePegoutId, usecases.TxBelowMinimumError)
-	}
 
 	useCase.rskWalletMutex.Lock()
```
Luisfc68: I think the mutex should be passed to the handler. Otherwise if you call the function from somewhere else or scan the file without context it can throw false positive alerts for possible race conditions

internal/usecases/pegout/bridge_pegout.go
```diff
@@ -20,115 +18,91 @@ const (
...
-	config := blockchain.NewTransactionConfig(totalValue, BridgeConversionGasLimit, entities.NewWei(BridgeConversionGasPrice))
-	receipt, txErr := useCase.rskWallet.SendRbtc(ctx, config, useCase.contracts.Bridge.GetAddress())
-	if txErr == nil {
-		log.Debugf("%s: transaction sent to the bridge successfully (%s)", usecases.BridgePegoutId, receipt.TransactionHash)
-	}
+	return useCase.handler.Execute(ctx, pegoutConfig, watchedQuotes)
+}
 
-	err = useCase.updateQuotes(ctx, receipt, txErr, watchedQuotes)
+func checkBalance(ctx context.Context, rskWallet blockchain.RootstockWallet, requiredBalance *entities.Wei) error {
```
Luisfc68: if you're going to covert this into regular functions to share accross the rebalance handlers, then put them in a separate package inside the pegout package, otherwise this is callable form the rest of use cases and it would be very confusing to have access to a function called `checkBalance` from a use case that has nothing to do with it. Same comment for the rest of the functions

internal/usecases/pegout/bridge_pegout.go
```diff
@@ -20,115 +18,91 @@ const (
...
-		if watchedQuote.PegoutQuote.Value != nil {
-			totalValue.Add(totalValue, watchedQuote.PegoutQuote.Value)
-		}
-		if watchedQuote.PegoutQuote.CallFee != nil {
-			totalValue.Add(totalValue, watchedQuote.PegoutQuote.CallFee)
-		}
-		if watchedQuote.PegoutQuote.GasFee != nil {
-			totalValue.Add(totalValue, watchedQuote.PegoutQuote.GasFee)
-		}
+func quoteContribution(wq quote.WatchedPegoutQuote) *entities.Wei {
```
Luisfc68: you already have the `Total()` function of the `PegoutQuote` struct.

internal/usecases/pegout/rebalance_all_at_once.go
```diff
@@ -0,0 +1,97 @@
...
+		log.Debugf("%s: transaction sent to the bridge successfully (%s)", usecases.BridgePegoutId, receipt.TransactionHash)
+	}
+
+	if err := updateQuotes(ctx, h.quoteRepository, receipt, txErr, watchedQuotes); err != nil {
+		return usecases.WrapUseCaseError(usecases.BridgePegoutId, err)
+	}
+	return nil
+}
+
+func updateQuotes(
```
Luisfc68: this function is specific to the handler it should be a receiver

internal/usecases/pegout/rebalance_all_at_once.go
Luisfc68, line 1: The handler should be individually tested.

internal/usecases/pegout/rebalance_utxo_split.go
```diff
@@ -0,0 +1,213 @@
...
+	if len(retained.BridgeRebalances) == 0 || retained.BridgeRefundTxHash != "" {
+		return
+	}
+	first := retained.BridgeRebalances[0]
+	retained.BridgeRefundTxHash = first.TxHash
+	retained.BridgeRefundGasUsed = first.GasUsed
+	retained.BridgeRefundGasPrice = first.GasPrice
+}
+
+func minWei(a, b *entities.Wei) *entities.Wei {
```
Luisfc68: This fucntion should be either a receiver of Wei or live in the same package as Wei

internal/usecases/pegout/rebalance_utxo_split.go
```diff
@@ -0,0 +1,213 @@
...
+		return
+	}
+	retained.BridgeRebalances = append(retained.BridgeRebalances, quote.BridgeRebalanceAllocation{
+		TxHash:   receipt.TransactionHash,
+		GasUsed:  receipt.GasUsed.Uint64(),
+		GasPrice: receipt.GasPrice,
+	})
+}
+
+func setDeprecatedRefundFields(retained *quote.RetainedPegoutQuote) {
```
Luisfc68: this is fine, but I don't see any modification to the DTO to mark these fields as deprecated

internal/usecases/pegout/rebalance_utxo_split.go
```diff
@@ -0,0 +1,213 @@
...
+}
+
+func getRemaining(wq quote.WatchedPegoutQuote) *entities.Wei {
+	if wq.RetainedQuote.RemainingToRefund != nil {
+		return wq.RetainedQuote.RemainingToRefund.Copy()
+	}
+	return quoteContribution(wq)
+}
+
+func appendAllocation(retained *quote.RetainedPegoutQuote, receipt blockchain.TransactionReceipt) {
```
Luisfc68: why this is not a receiver?

Luisfc68: also its more clear if its called `AppendRebalanceAllocation`

internal/usecases/pegout/rebalance_utxo_split.go
```diff
@@ -0,0 +1,213 @@
...
+
+func adjustTotalForRetries(watchedQuotes []quote.WatchedPegoutQuote) *entities.Wei {
+	adjusted := entities.NewWei(0)
+	for _, wq := range watchedQuotes {
+		adjusted.Add(adjusted, getRemaining(wq))
+	}
+	return adjusted
+}
+
+func getRemaining(wq quote.WatchedPegoutQuote) *entities.Wei {
```
Luisfc68: this can be also implemented as a receiver

internal/usecases/pegout/rebalance_utxo_split.go
```diff
@@ -0,0 +1,213 @@
...
+func buildChunkAmounts(numTxs uint64, bridgeMin, remainder *entities.Wei) []*entities.Wei {
+	chunkAmounts := make([]*entities.Wei, numTxs)
+	chunkAmounts[0] = new(entities.Wei).Add(bridgeMin.Copy(), remainder)
+	for i := uint64(1); i < numTxs; i++ {
+		chunkAmounts[i] = bridgeMin.Copy()
+	}
+	return chunkAmounts
+}
+
+func adjustTotalForRetries(watchedQuotes []quote.WatchedPegoutQuote) *entities.Wei {
```
Luisfc68: this is specific to the handler, it should be a receiver

internal/usecases/pegout/rebalance_utxo_split.go
Luisfc68, line 1: as a general comment, all the functions that are specific to the handler's logic should be receivers, otherwise put them in a file where we can distinguish them as utilities or similar

internal/usecases/pegout/rebalance_utxo_split.go
```diff
@@ -0,0 +1,213 @@
...
+	}
+	return h.distributeChunkReceipt(ctx, chunkIndex, totalChunks, receipt, watchedQuotes)
+}
+
+func isValidSplitReceipt(receipt blockchain.TransactionReceipt, chunkIndex, totalChunks int) bool {
+	if receipt.TransactionHash == "" {
+		log.Errorf("%s: split tx %d/%d failed: incomplete receipt", usecases.BridgePegoutId, chunkIndex+1, totalChunks)
+		return false
+	}
+	log.Debugf("%s: split tx %d/%d sent to the bridge successfully (%s)", usecases.BridgePegoutId, chunkIndex+1, totalChunks, receipt.TransactionHash)
```
Luisfc68: the success log shouldn't be before the last validation

internal/usecases/pegout/rebalance_utxo_split.go
```diff
@@ -0,0 +1,213 @@
...
+		log.Errorf("%s: split tx %d/%d failed: %v", usecases.BridgePegoutId, chunkIndex+1, totalChunks, txErr)
+		return watchedQuotes, nil
+	}
+	if !isValidSplitReceipt(receipt, chunkIndex, totalChunks) {
+		return watchedQuotes, nil
+	}
+	return h.distributeChunkReceipt(ctx, chunkIndex, totalChunks, receipt, watchedQuotes)
+}
+
+func isValidSplitReceipt(receipt blockchain.TransactionReceipt, chunkIndex, totalChunks int) bool {
```
Luisfc68: shouldn't this function include a status check?

internal/usecases/pegout/rebalance_utxo_split.go
```diff
@@ -0,0 +1,213 @@
...
+	}
+	log.Debugf("%s: split tx %d/%d sent to the bridge successfully (%s)", usecases.BridgePegoutId, chunkIndex+1, totalChunks, receipt.TransactionHash)
+	if receipt.Value == nil {
+		log.Errorf("%s: split tx %d/%d failed: missing receipt value", usecases.BridgePegoutId, chunkIndex+1, totalChunks)
+		return false
+	}
+	return true
+}
+
+func (h *UtxoSplitHandler) distributeChunkReceipt(
```
Luisfc68, line 128: I think this function should be either split or refactored, mainly because of the way that the `for` loop is implemented, if you're going to define a `for` with a condition (like `while` style), maybe you should design the loop so your logic doesn't rely on `continue` and `break` as both keywords degrade the readability of the function

internal/usecases/pegout/rebalance_utxo_split.go
```diff
@@ -0,0 +1,213 @@
...
+		taken := minWei(quoteRemaining, chunkRemaining)
+		quoteRemaining.Sub(quoteRemaining, taken)
+		chunkRemaining.Sub(chunkRemaining, taken)
+		if quoteRemaining.Cmp(zero) == 0 {
+			watchedQuote.RetainedQuote.RemainingToRefund = entities.NewWei(0)
+			watchedQuote.RetainedQuote.State = quote.PegoutStateBridgeTxSucceeded
+		} else {
+			watchedQuote.RetainedQuote.RemainingToRefund = quoteRemaining.Copy()
+		}
+		setDeprecatedRefundFields(&watchedQuote.RetainedQuote)
```
Luisfc68: this is being called multiple times but it needs to happen in all the cases, maybe you could summarize in just one invocation earlier in the function

internal/usecases/pegout/rebalance_utxo_split.go
Luisfc68: also can be a receiver

internal/usecases/pegout/rebalance_all_at_once.go
```diff
@@ -0,0 +1,97 @@
...
+		contracts:       contracts,
+	}
+}
+
+func (h *AllAtOnceHandler) Execute(
+	ctx context.Context,
+	pegoutConfig liquidity_provider.PegoutConfiguration,
+	watchedQuotes []quote.WatchedPegoutQuote,
+) error {
+	totalValue := calculateTotalToPegout(watchedQuotes)
```
Luisfc68: what happens if I have the other strategy and some quote is partially refunded? if I recall correctly from the other file the state only changes when is completely refunded so this would include the total value of that quote, I think here you should sum the remaining to refund amounts of each quote

internal/usecases/pegout/bridge_pegout_test.go
Luisfc68, line 1: The handler should be individually tested.

internal/adapters/dataproviders/database/mongo/migrations/runner.go
Luisfc68, line 1: If I saw correctly, only RunAll is used from this file, not sure if we should add and maintain code we don't intent to use

internal/adapters/dataproviders/database/mongo/migrations/scripts/000002_bridge_rebalances_tx_hash_index.up.json
Luisfc68, line 1: I think the index creation should continue as it currently is (done programatically in createIndexes)

internal/adapters/dataproviders/database/mongo/migrations/scripts/000002_bridge_rebalances_tx_hash_index.down.json
Luisfc68, line 1: same comment in general about the down migrations, if we don't plan to use them we shouldn't add them without a reason

Luisfc68 (changes requested):

internal/entities/quote/pegout_quote.go
volodymyrzahanych: Comment added 
```
// RemainingToRefund tracks how much of the quote's total value (Value + CallFee + GasFee) still needs
// to be covered by bridge refund transactions. It is initialized to quote.Total() when the quote is
// accepted and decremented as each bridge refund transaction covers a portion of the amount. When it
// reaches zero the quote transitions to PegoutStateBridgeTxSucceeded. A nil value is treated as
// quote.Total() for backwards compatibility with records created before this field was introduced.
```

volodymyrzahanych: quote.Total() is also handled

internal/entities/quote/pegout_quote.go
volodymyrzahanych: omitempty is deleted

internal/configuration/registry/usecase.go
volodymyrzahanych: the signature is updated

internal/usecases/pegout/bridge_pegout.go
volodymyrzahanych: bridge_pegout.go — removed rskWalletMutex from BridgePegoutUseCase and its Run(). Added mutex sync.Locker param to NewRebalanceHandler, forwarding it to both handler constructors.
rebalance_all_at_once.go / rebalance_utxo_split.go — each handler now owns the mutex, locking at the top of Execute() so the wallet operations are self-contained and the concurrency intent is visible where the wallet is actually used.
usecase.go — mutex now passed to NewRebalanceHandler instead of NewBridgePegoutUseCase.
bridge_pegout_test.go / pegout_bridge_watcher_test.go — all call sites updated accordingly.

volodymyrzahanych: bridge_pegout.go — removed rskWalletMutex from BridgePegoutUseCase and its Run(). Added mutex sync.Locker param to NewRebalanceHandler, forwarding it to both handler constructors.
rebalance_all_at_once.go / rebalance_utxo_split.go — each handler now owns the mutex, locking at the top of Execute() so the wallet operations are self-contained and the concurrency intent is visible where the wallet is actually used.
usecase.go — mutex now passed to NewRebalanceHandler instead of NewBridgePegoutUseCase.
bridge_pegout_test.go / pegout_bridge_watcher_test.go — all call sites updated accordingly.

internal/usecases/pegout/bridge_pegout.go
volodymyrzahanych: QuoteContribution now delegates to wq.PegoutQuote.Total() instead of duplicating its logic.

internal/usecases/pegout/rebalance_all_at_once.go
volodymyrzahanych: this function is refactored to be a receiver

volodymyrzahanych: this function is refactored to be a receiver

internal/usecases/pegout/rebalance_all_at_once.go
volodymyrzahanych: The new file [rebalance_all_at_once_test.go](vscode-webview://04j4imsh71febdiea1enflumt87gm1rr7dds9q9k7bk1rbhqhjiq/internal/usecases/pegout/rebalance_all_at_once_test.go) tests AllAtOnceHandler.Execute directly, covering:

success — tx and update both succeed; all quotes get BridgeTxSucceeded state with tx hash fields populated
total below minimum — returns TxBelowMinimumError with no wallet/repo calls
balance check error — GetBalance fails; no send or update
insufficient balance — returns InsufficientAmountError; no send or update
bridge tx fails — quotes updated with BridgeTxFailed state and empty hash fields; tx error returned
quotes update fails — tx succeeded but update fails; update error returned
bridge tx fails and quotes update fails — both errors are joined and returned

internal/usecases/pegout/rebalance_utxo_split.go
volodymyrzahanych: minWei is now (*Wei).Min in [internal/entities/wei.go](vscode-webview://04j4imsh71febdiea1enflumt87gm1rr7dds9q9k7bk1rbhqhjiq/internal/entities/wei.go), consistent with Add/Sub/Mul/Div — receiver holds the result.

internal/usecases/pegout/rebalance_utxo_split.go
volodymyrzahanych: done

internal/usecases/pegout/rebalance_utxo_split.go
volodymyrzahanych: done

internal/usecases/pegout/rebalance_utxo_split.go
volodymyrzahanych: done

volodymyrzahanych: done

internal/usecases/pegout/rebalance_utxo_split.go
volodymyrzahanych: done

internal/usecases/pegout/rebalance_utxo_split.go
volodymyrzahanych: done

volodymyrzahanych: done

internal/usecases/pegout/rebalance_utxo_split.go
volodymyrzahanych: done

internal/usecases/pegout/rebalance_utxo_split.go
volodymyrzahanych: done

internal/usecases/pegout/rebalance_utxo_split.go
volodymyrzahanych: Solution: Split the function into four focused pieces:

distributeChunkReceipt — now the orchestrator. It flushes zero-remaining quotes first, then iterates through quotes consuming the chunk. The loop body has no continue or break; when a quote is only partially covered, it prepends it back and returns early (single exit point within the loop).

flushFullyCoveredQuotes — drains quotes from the front of the slice that already have zero remaining. This was the first if/continue block that was unrelated to the chunk allocation itself.

allocateChunkToQuote — pure computation: takes the min of chunk remaining and quote remaining, subtracts, sets state. No persistence, no slice manipulation.

persistQuoteUpdate — wraps the repository call with the error logging. Eliminates the duplicated error-handling code that appeared twice in the original loop.

The loop in distributeChunkReceipt now reads linearly: pop a quote, allocate, persist, and if it's not fully covered, put it back and stop.

internal/usecases/pegout/rebalance_utxo_split.go
volodymyrzahanych: Now SetDeprecatedRefundFields() is called exactly once, at line 201 inside persistQuoteUpdate — the single gateway to the database. Both flushFullyCoveredQuotes and allocateChunkToQuote no longer need to worry about it; it's guaranteed to happen right before the persist regardless of which code path produced the quote update.

internal/usecases/pegout/rebalance_all_at_once.go
volodymyrzahanych: common/helpers.go — QuoteContribution: Changed from wq.PegoutQuote.Total() to wq.Remaining(). This is the core of your catch — if a quote was partially refunded by a prior UTXO split attempt (which failed before completing all chunks), the state wouldn't have changed to BridgeTxSucceeded, so the quote would still be in the watched set. Using Total() would overcount the amount to send to the bridge.

rebalance_all_at_once.go — updateQuotes allocation tracking: Replaced the manual BridgeRebalances = []BridgeRebalanceAllocation{...} assignment with AppendRebalanceAllocation, which preserves any allocation records from a prior partial UTXO split attempt. Also added SetDeprecatedRefundFields() for consistency with the UTXO split handler.

rebalance_all_at_once.go — RemainingToRefund on success: Added RemainingToRefund = entities.NewWei(0) when the transaction succeeds. Without this, Remaining() would still return the old value on any subsequent read (e.g., if the quote were re-fetched before the state transition is consumed).

internal/adapters/dataproviders/database/mongo/migrations/runner.go
volodymyrzahanych: done

internal/usecases/pegout/bridge_pegout_test.go
volodymyrzahanych: done

internal/adapters/dataproviders/database/mongo/migrations/scripts/000002_bridge_rebalances_tx_hash_index.up.json
volodymyrzahanych: Summary of changes:

Deleted 000002_bridge_rebalances_tx_hash_index.{up,down}.json
Added bridge_rebalances.tx_hash non-unique index creation to createIndexes in [mongo.go](vscode-webview://04j4imsh71febdiea1enflumt87gm1rr7dds9q9k7bk1rbhqhjiq/internal/adapters/dataproviders/database/mongo/mongo.go) via a new createIndex helper
Removed all tests in runner_test.go that relied on deleted functions (Down, MigrateTo, findVersion, removeVersion) and updated version expectations to reflect only 1 migration script remaining

volodymyrzahanych: Summary of changes:

Deleted 000002_bridge_rebalances_tx_hash_index.{up,down}.json
Added bridge_rebalances.tx_hash non-unique index creation to createIndexes in [mongo.go](vscode-webview://04j4imsh71febdiea1enflumt87gm1rr7dds9q9k7bk1rbhqhjiq/internal/adapters/dataproviders/database/mongo/mongo.go) via a new createIndex helper
Removed all tests in runner_test.go that relied on deleted functions (Down, MigrateTo, findVersion, removeVersion) and updated version expectations to reflect only 1 migration script remaining

volodymyrzahanych: Deleted the .down.json file and removed the now-unnecessary migrationSuffixes map and matchSuffix helper — assignMigrationFile directly checks for .up.json.

internal/adapters/dataproviders/rootstock/wallet.go
```diff
@@ -145,24 +145,19 @@ func (wallet *RskWalletImpl) buildTransactionReceipt(receipt *geth.Receipt, tx *
...
 		To:                toAddressStr,
+		Status:            receipt.Status == geth.ReceiptStatusSuccessful,
 		CumulativeGasUsed: new(big.Int).SetUint64(receipt.CumulativeGasUsed),
 		GasUsed:           new(big.Int).SetUint64(receipt.GasUsed),
 		Value:             txValue,
 		GasPrice:          entities.NewWei(receipt.EffectiveGasPrice.Int64()),
 		Logs:              make([]blockchain.TransactionLog, 0),
 	}
-
-	if receipt.Status == 0 {
```
Luisfc68, line 161: I think that in order to minimize the impact of this task in the rest of the application, we should keep returning an error if `Status == 0`. In your particular case where you need to know the value of `Status` you can wrap a sentinel error here (e.g. TxFailedError) and check that outside with `errors.Is`. In that way you don't need to worry about introducing potential issues into preexistent consumers of this function

Luisfc68 (changes requested):

internal/adapters/dataproviders/rootstock/wallet.go
volodymyrzahanych: done

Luisfc68: The idea of this suggestion was to avoid touching already existing structures. You added the recommendation, but it wasn't just about updating the code, was about reducing the impact in the existing codebase. For example, the new field you added in `blockchain.TransactionReceipt` was populated only in the places required for the task, but you ignored the rest of the initializations of the structure (the ones present in the contract files), this may lead to potential issues on the future, its very important that if you change any signature, struct, etc. Consider the impact of those changes in the overall code, not just the part related to the task. If this is not possible or requires more effort (which is understandable) you should take a cleaner approach like the one suggested in the comment, adding the new sentinel error removes the need of modifying the struct, therefore we don't run the risk of forgetting to update any initialization.

Luisfc68: The main idea is, when extending the code, avoid affecting the existing one as much as possible

volodymyrzahanych: done

Luisfc68 (approved):

PR #937, opened 2026-03-18
Feat: Fly2241 btc rbtc ratio management

AndresQuijano:
## What
This is the second part of the feature to make the BTC/RBTC liquidity ratio configurable instead of keep it as 50-50

## Why
Give the chance to split the liquidity in a more convenient way for the liquidity provider

## Type of Change
- [ ] Breaking change 
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test updates
- [ ] Security fix

- [x] Management UI / API
- [ ] PegIn flow
- [ ] PegOut flow
- [ ] Utility scripts
- [ ] Configuration files
- [ ] Metrics and alerting

## Related Issues
[Jira ticket](https://rsklabs.atlassian.net/browse/FLY-2241)

## How to test
Start the LPS in local and do the magic with the password.
Authenticate in the UI management.
Go to the next card for liquidity ratio and move the values around.
Check that, after an update, the cooldown gets activated

## Reviewer Guidelines
The ratio was initially a configuration variable but with this PR it is moved to be a value that can be only updated via API
The not firing condition during cooldown was removed from low liquidity alerts by product manager recommendation

image


internal/usecases/liquidity_provider/get_liquidity_ratio.go
```diff
@@ -0,0 +1,150 @@
...
+		return LiquidityRatioDetail{}, usecases.WrapUseCaseError(usecases.GetLiquidityRatioId, err)
+	}
+
+	isPreview := proposedBtcPercentage > 0
+	btcPercentage := stateConfig.BtcLiquidityTargetPercentage
+	if isPreview {
+		btcPercentage = proposedBtcPercentage
+	}
+	rbtcPercentage := 100 - btcPercentage
+
```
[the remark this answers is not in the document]
AndresQuijano: For this app we run this validations in handlers.

internal/adapters/entrypoints/rest/handlers/get_liquidity_ratio.go
```diff
@@ -0,0 +1,45 @@
...
+	return func(w http.ResponseWriter, req *http.Request) {
+		var proposedPercentage uint64
+		if param := req.URL.Query().Get("btcPercentage"); param != "" {
+			parsed, err := strconv.ParseUint(param, 10, 64)
+			if err != nil {
+				rest.ValidateRequestError(w, err)
+				return
+			}
+			proposedPercentage = parsed
+		}
```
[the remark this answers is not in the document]
AndresQuijano: Changed.

internal/usecases/liquidity_provider/set_liquidity_ratio.go
```diff
@@ -0,0 +1,56 @@
...
+		return usecases.WrapUseCaseError(usecases.SetLiquidityRatioId, err)
+	}
+
+	if stateConfig.BtcLiquidityTargetPercentage == btcPercentage {
+		return nil
+	}
+
+	stateConfig.BtcLiquidityTargetPercentage = btcPercentage
+	stateConfig.RatioCooldownEndTimestamp = time.Now().Unix() + CooldownAfterRatioChange
+
```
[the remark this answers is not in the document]
AndresQuijano: Already validated in hanlder

internal/usecases/liquidity_provider/set_liquidity_ratio_test.go
```diff
@@ -0,0 +1,121 @@
...
+		RatioCooldownEndTimestamp:    0,
+	}
+	provider.On("StateConfiguration", ctx).Return(stateConfig, nil)
+	walletMock.On("SignBytes", mock.Anything).Return([]byte{1, 2, 3}, nil)
+
+	now := time.Now().Unix()
+	lpRepository.On("UpsertStateConfiguration", ctx, mock.MatchedBy(func(config entities.Signed[lpEntity.StateConfiguration]) bool {
+		return config.Value.BtcLiquidityTargetPercentage == 70 &&
+			config.Value.RatioCooldownEndTimestamp >= now+liquidity_provider.CooldownAfterRatioChange-1 &&
+			config.Value.RatioCooldownEndTimestamp <= now+liquidity_provider.CooldownAfterRatioChange+1
```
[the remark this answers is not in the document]
AndresQuijano: Changed

OpenApi.yml
```diff
@@ -293,6 +309,48 @@ components:
       - status
       - providerType
       type: object
+    LiquidityRatioResponse:
+      properties:
+        btcCurrentBalance:
+          $ref: '#/components/schemas/'
+        btcImpact:
```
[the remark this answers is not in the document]
AndresQuijano: Noise, this is how that type of data is handled across the file

internal/usecases/liquidity_provider/get_liquidity_ratio.go
```diff
@@ -0,0 +1,150 @@
...
+		BtcImpact:               calculateImpact(btcBalance, btcTarget, btcThreshold),
+		RbtcImpact:              calculateImpact(rbtcBalance, rbtcTarget, rbtcThreshold),
+		CooldownActive:          time.Now().Unix() < stateConfig.RatioCooldownEndTimestamp,
+		CooldownEndTimestamp:    stateConfig.RatioCooldownEndTimestamp,
+		CooldownDurationSeconds: CooldownAfterRatioChange,
+		IsPreview:               isPreview,
+	}, nil
+}
+
+func calculateTarget(maxLiquidity *entities.Wei, percentage uint64) (*entities.Wei, error) {
```
Luisfc68: Pls if the function is specific to the use case make it a receiver, even if it doesn't use the state of the use case, the reason of this is because otherwise it will be accesible from other use cases in the package, which can lead to errors in the future, if we move to use one use case per package in the future, this would be ok, but in the current set up I think it's not worth

Luisfc68: same comment for the functions below

internal/usecases/liquidity_provider/get_liquidity_ratio.go
```diff
@@ -0,0 +1,150 @@
...
+	return new(entities.Wei).Div(
+		new(entities.Wei).Mul(maxLiquidity, entities.NewUWei(percentage)),
+		entities.NewUWei(100),
+	)
+}
+
+func calculateThreshold(target *entities.Wei, tolerance liquidity_provider.ExcessTolerance) *entities.Wei {
```
Luisfc68: isn't this very similar to the one you used in the cold wallet use case? if yes you could create a separate package with utilities (doesn't need to be in this PR)

internal/usecases/liquidity_provider/initialize_state_configuration.go
```diff
@@ -13,15 +13,17 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
-const CooldownAfterRatioChange int64 = 10800 // 3 hours
+const (
+	CooldownAfterRatioChange            int64  = 10800 // 3 hours
+	DefaultBtcLiquidityTargetPercentage uint64 = 50
```
Luisfc68: shouldn't this be in the defaults file?

pkg/liquidity_provider_test.go
Luisfc68, line 1: you created a separate file called liquidity_ratio, I think you should do the same for the tests

Luisfc68 (changes requested):

internal/usecases/liquidity_provider/get_liquidity_ratio.go
AndresQuijano: Yes, good catch, refactored

internal/usecases/liquidity_provider/initialize_state_configuration.go
AndresQuijano: I moved DefaultBtcLiquidityTargetPercentage to the default file but I think CooldownAfterRatioChange should live here

internal/usecases/liquidity_provider/get_liquidity_ratio.go
AndresQuijano: Done

pkg/liquidity_provider_test.go
AndresQuijano: I moved the tests to a separated file

internal/usecases/liquidity_provider/initialize_state_configuration.go
Luisfc68: Yes, I agree

Luisfc68 (approved):

PR #940, opened 2026-03-23
Fix: Input validation in recommended pegin and pegout

AndresQuijano:
## What
Fix three bugs related to validation in recommended pegin and pegout

## Why
Improve quality

## Type of Change
- [ ] Breaking change 
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test updates
- [ ] Security fix

- [ ] Management UI / API
- [x] PegIn flow
- [x] PegOut flow
- [ ] Utility scripts
- [ ] Configuration files
- [ ] Metrics and alerting

## Related Issues
[FLY-2220](https://rsklabs.atlassian.net/browse/FLY-2220)
[FLY-2221](https://rsklabs.atlassian.net/browse/FLY-2221)
[FLY-2222](https://rsklabs.atlassian.net/browse/FLY-2222)

## How to test
Follow the instructions in the tickets

### DIFF
--- a/internal/adapters/entrypoints/rest/handlers/recommended_pegin.go
+++ b/internal/adapters/entrypoints/rest/handlers/recommended_pegin.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"github.com/rsksmart/liquidity-provider-server/internal/adapters/entrypoints/rest"
 	"github.com/rsksmart/liquidity-provider-server/internal/entities"
+	"github.com/rsksmart/liquidity-provider-server/internal/entities/blockchain"
 	"github.com/rsksmart/liquidity-provider-server/internal/entities/liquidity_provider"
 	"github.com/rsksmart/liquidity-provider-server/internal/usecases"
 	"github.com/rsksmart/liquidity-provider-server/pkg"
@@ -41,6 +42,17 @@ func NewRecommendedPeginHandler(useCase RecommendedPeginUseCase) http.HandlerFun
 		parsedAmount, ok := new(big.Int).SetString(amount, 10)
 		if !ok {
 			jsonErr := rest.NewErrorResponse("invalid or missing parameter amount", true)
+			rest.JsonErrorResponse(w, http.StatusBadRequest, jsonErr)
+			return
+		}
+		if parsedAmount.Sign() <= 0 {
+			jsonErr := rest.NewErrorResponse("parameter amount must be greater than zero", true)
+			rest.JsonErrorResponse(w, http.StatusBadRequest, jsonErr)
+			return
+		}
+
+		if destinationAddress != "" && !blockchain.IsRskAddress(destinationAddress) {
+			jsonErr := rest.NewErrorResponse("invalid parameter destination_address", true)
 			rest.JsonErrorResponse(w, http.StatusBadRequest, jsonErr)
 			return
 		}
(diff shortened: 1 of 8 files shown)


internal/usecases/pegin/recommended_pegin.go
```diff
@@ -41,6 +41,12 @@ func (useCase *RecommendedPeginUseCase) Run(
 	config := useCase.peginProvider.PeginConfiguration(ctx)
 	result := new(big.Int).Set(userBalance.AsBigInt())
 
+	if userBalance.Cmp(config.MinValue) < 0 {
+		outOfRangeErr := fmt.Errorf("%w [%v, %v]", liquidity_provider.AmountOutOfRangeError, config.MinValue, config.MaxValue)
+		err := fmt.Errorf("provided amount %s is out of range: %w", userBalance.String(), outOfRangeErr)
```
Hakob23: nit: Is there a specific reason, that we have separate "provided amount %s is out of range:" and "parameter amount must be greater than zero" error messages? Aren't they equivalent? If yes, then may be we could carry on with a single error message?

internal/adapters/entrypoints/rest/handlers/recommended_pegin.go
```diff
@@ -44,6 +45,17 @@ func NewRecommendedPeginHandler(useCase RecommendedPeginUseCase) http.HandlerFun
 			rest.JsonErrorResponse(w, http.StatusBadRequest, jsonErr)
 			return
 		}
+		if parsedAmount.Sign() <= 0 {
+			jsonErr := rest.NewErrorResponse("parameter amount must be greater than zero", true)
```
Hakob23, line 49: This is the same validation logic, as in recommended_pegout, may be could have a small shared validation logic module for both handlers?

internal/usecases/pegin/recommended_pegin.go
```diff
@@ -41,6 +41,12 @@ func (useCase *RecommendedPeginUseCase) Run(
 	config := useCase.peginProvider.PeginConfiguration(ctx)
 	result := new(big.Int).Set(userBalance.AsBigInt())
 
+	if userBalance.Cmp(config.MinValue) < 0 {
```
Hakob23: same suggestion of having a shared validation logic(but this time with  `if err := config.ValidateAmount(userBalance)` from recommended_pegout, as the logic is essentially the same)

Hakob23 (approved):
I have added two small refactoring suggestions for the validation logic, but I don't have hard position on its implementation. Apart from that, LGTM!

internal/usecases/pegin/recommended_pegin.go
AndresQuijano: Thansk, I'm aligning this one to the ValidateAmount standard of the rest of the similar use cases.

internal/adapters/entrypoints/rest/handlers/recommended_pegin.go
AndresQuijano: Agreed it's duplicated, but with only 2 handlers, a shared helper adds indirection without meaningful gain.

internal/usecases/pegin/recommended_pegin.go
AndresQuijano: After the fix in comment 1, both usecases now use config.ValidateAmount consistently, which was the real inconsistency. Extracting further would require a cross-package helper with 3 parameters to abstract 3 lines — the abstraction complexity matches the code itself. Keeping it inline.

PR #941, opened 2026-03-25
feat: add Bitcoin/Rootstock reorg monitoring

volodymyrzahanych:
Introduce node reorg watchers with shared depth-check use case, Prometheus metrics, and env-configured thresholds.

## What
Added BitcoinReorgWatcher and RootstockReorgWatcher to continuously detect chain reorgs by tracking recent block hashes in-memory.
Made max reorg depth configurable via environment variable (defaults to 2 when unset) and retains a buffer of recent hashes (targeting (N \gg depth), e.g. depth + 10) to support detection.
Triggers an alert when a detected reorg exceeds the configured depth, and added tests covering watcher behavior and configuration defaults.

## Why

## Type of Change
- [ ] Breaking change 
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test updates
- [ ] Security fix

- [ ] Management UI / API
- [ ] PegIn flow
- [ ] PegOut flow
- [ ] Utility scripts
- [ ] Configuration files
- [x] Metrics and alerting

## Related Issues
[Jira ticket](https://rsklabs.atlassian.net/browse/FLY-2244)

## How to test
- From the repository root, run `make test` (full suite with race detector and coverage as defined in the Makefile).
- For a quicker pass limited to this feature, run:

  `go test -race -timeout 30m ./internal/usecases/watcher/... ./internal/adapters/entrypoints/watcher/...`

- No extra services or local stacks are required for the above; they are standard Go unit tests.

Reorg monitoring exposes **five Prometheus series**, each labeled with **`node="bitcoin"`** or **`node="rootstock"`** (published on the LPS Prometheus scrape path alongside the other app metrics).

- **`lps_node_reorg_depth`** (gauge) — last detected reorg depth for that chain.
- **`lps_node_reorg_max_depth_threshold`** (gauge) — configured max depth before treating a reorg as severe (from `BITCOIN_MAX_REORG_DEPTH` / `RSK_MAX_REORG_DEPTH`, default **2** when unset).
- **`lps_node_reorg_above_threshold`** (gauge) — **1** if current depth exceeds the threshold, **0** otherwise.
- **`lps_node_reorg_check_errors_total`** (counter) — increments when a reorg check fails (e.g. RPC error) for that node.
- **`lps_node_reorg_alerts_total`** (counter) — increments when an alert is emitted for a reorg above the threshold for that node.

**How QA can validate:** confirm these names and label values appear in the scrape output in a running deployment, and that gauges/counters move in line with reorg-check events (depth/threshold/above flag stay consistent; errors and alerts only increase when the corresponding conditions occur). Automated coverage of the metrics wiring is in `internal/adapters/entrypoints/watcher/monitoring` tests (e.g. `TestReorgMetricsWatcher_StartUpdatesMetricsFromEvents`).

## Reviewer Guidelines

- **Architecture:** Detection lives in **`NodeReorgCheckUseCase`** (shared for BTC and RSK). The two entrypoint watchers only tick and call `Run` with the right `NodeType`. **`ReorgMetricsWatcher`** subscribes to the event bus and maps reorg events to Prometheus; keep that separation in mind when tracing behavior end-to-end.
- **In-memory state:** Per-node history is **not persisted**. After a process restart, the first successful tick re-seeds the window; reviewers should be comfortable with that trade-off for this feature.
- **Depth vs window:** History length is `reorgHistoryWindow(maxDepth)` (`maxDepth + 10`, minimum **12**). If the common ancestor is **outside** that window, depth is treated conservatively as **above threshold** (algorithm returns `maxDepth + 1`). Validate that edge case is acceptable.
- **Alerts:** Sent only when **`reorgDepth > maxDepth`** (strict inequality), with **`NODE_REORG_ALERT_COOLDOWN_SECONDS`** between alerts per node. Failed RPC paths increment **`lps_node_reorg_check_errors_total`** and return a wrapped use-case error from the tick.
- **Prometheus semantics:** **`lps_node_reorg_depth`** and **`lps_node_reorg_above_threshold`** reflect the **latest check** and can return to “no reorg” after the chain stabilizes. For external alerting on reorgs, prefer rules based on **`increase(lps_node_reorg_alerts_total[...])`** (or equivalent) rather than short-lived gauge spikes alone.
- **Configuration:** **`BITCOIN_MAX_REORG_DEPTH`** and **`RSK_MAX_REORG_DEPTH`** are independent; defaults are applied when unset (see `FillWithDefaults()` in environment config).

## Additional Notes


internal/adapters/dataproviders/bitcoin/mempool_space/api.go
Luisfc68, line 1: mempool space API does have the functionality to implement these functions, but I understand they're not needed for this task, however I recommend changing the error message to just "unimplemented" or something similar, because reorgs can be detected with external APIs as well

internal/adapters/dataproviders/bitcoin/rpc.go
Luisfc68, line 1: this changes are untested

internal/adapters/entrypoints/watcher/monitoring/metrics.go
```diff
@@ -7,13 +7,46 @@ import (
...
+	}, []string{"node"})
+	alerts = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Name: "lps_node_reorg_alerts_total",
+		Help: "Total number of reorganization alerts sent",
+	}, []string{"node"})
+	return
 }
 
 func NewMetrics(reg prometheus.Registerer) *Metrics {
+	reorgDepth, reorgMax, reorgAbove, reorgErrs, reorgAlerts := newNodeReorgMetrics()
```
Luisfc68: small tip:
in this case of situations prefer using a struct instead of multiple returns, because if two values have the same type (e.g. reorgDepth and reorgMax) the caller my switch them by mistake and it would be hard to detect

internal/adapters/entrypoints/watcher/common.go
```diff
@@ -21,6 +21,8 @@ const (
 	rskEclipseCheckInterval          = 15 * time.Second
 	btcReleaseCheckInterval          = 3 * time.Minute
 	assetMetricsUpdateInterval       = 1 * time.Minute
+	bitcoinReorgCheckInterval        = 1 * time.Minute
```
Luisfc68: maybe we can set these intervals to the respective block times (30s rsk and 10m btc). WDYT?

internal/usecases/watcher/node_reorg_check.go
```diff
@@ -0,0 +1,434 @@
...
+			entities.NodeTypeBitcoin:   bitcoinMaxReorgDepth,
+			entities.NodeTypeRootstock: rootstockMaxReorgDepth,
+		},
+		alertCooldown:   alertCooldown,
+		lastAlertByNode: make(map[entities.NodeType]time.Time),
+	}
+}
+
+func reorgHistoryWindow(maxDepth uint64) uint64 {
+	window := maxDepth + 10
```
Luisfc68: Please name the 10 and the 12, maybe the names could be
- minReorgWindow (12)
- reorgWindowSize (10)

internal/usecases/watcher/node_reorg_check.go
```diff
@@ -0,0 +1,434 @@
...
+	useCase.lastAlertByNode[nodeType] = time.Now()
+	useCase.eventBus.Publish(blockchain.NodeReorgAlertSentEvent{
+		BaseEvent:     entities.NewBaseEvent(blockchain.NodeReorgAlertSentEventId),
+		NodeType:      nodeType,
+		DetectedDepth: reorgDepth,
+	})
+	return nil
+}
+
+func (useCase *NodeReorgCheckUseCase) runBitcoin(ctx context.Context) error {
```
Luisfc68, line 127: please prefer descriptive names over abbreviations:
- window or historyWindow >>> hw
- state or btcState or bitcoinState >>> st 
- header >>> hdr

internal/usecases/watcher/node_reorg_check.go
```diff
@@ -0,0 +1,434 @@
...
+	tipHeight uint64,
+	maxDepth uint64,
+	hw uint64,
+) error {
+	if err := refreshBitcoinWindow(st, btc, tipHeight, hw); err != nil {
+		return useCase.reorgError(entities.NodeTypeBitcoin, err)
+	}
+	st.lastTipHash = tipHash
+	st.lastTipHeight = tipHeight
+	useCase.publishReorgCheck(entities.NodeTypeBitcoin, 0, maxDepth, false)
```
Luisfc68: I'd call this outside of the function as it's not part of the initialization

internal/usecases/watcher/node_reorg_check.go
```diff
@@ -0,0 +1,434 @@
...
+	tipHash string,
+	tipHeight uint64,
+	maxDepth uint64,
+	hw uint64,
+) error {
+	st.lastTipHash = tipHash
+	st.lastTipHeight = tipHeight
+	if err := refreshBitcoinWindow(st, btc, tipHeight, hw); err != nil {
+		return useCase.reorgError(entities.NodeTypeBitcoin, err)
+	}
```
Luisfc68: same here, its not part of the advancement, you should call this function outside

internal/usecases/watcher/node_reorg_check.go
```diff
@@ -0,0 +1,434 @@
...
+	parentHash, err := useCase.getBitcoinParentHash(btc, tipHash)
+	if err != nil {
+		return err
+	}
+
+	if parentHash == st.lastTipHash {
+		return useCase.advanceBitcoinTip(st, btc, tipHash, tipHeight, maxDepth, hw)
+	}
+
+	reorgDepth := bitcoinReorgDepth(btc, st, tipHeight, hw, maxDepth)
```
Luisfc68: the name `bitcoinReorgDepth` is confusing because it looks like you assume that this is already a reorg but maybe there was more than 2 blocks mined between use case executions

internal/usecases/watcher/node_reorg_check.go
```diff
@@ -0,0 +1,434 @@
...
+			return st.lastTipHeight - h
+		}
+		if h == 0 {
+			return maxDepth + 1
+		}
+		h--
+	}
+}
+
+func (useCase *NodeReorgCheckUseCase) runRootstock(ctx context.Context) error {
```
Luisfc68, line 336: my comments here are the same I made for the btc part

internal/usecases/watcher/node_reorg_check.go
```diff
@@ -0,0 +1,434 @@
...
+		hash, err := btc.GetBlockHashAtHeight(int64(h))
+		if err != nil {
+			return err
+		}
+		st.knownHashByHeight[h] = hash
+	}
+	return nil
+}
+
+func bitcoinReorgDepth(
```
Luisfc68: I think there is a correctness issue here. Reaching this branch does not necessarily imply a reorg: it can also happen if more than one block was mined between executions. In that case, this function is not detecting an actual reorg depth, only the distance to the last common block found in the local history window. If the chain advanced by more than the configured history window, this can raise a false positive reorg alert

internal/usecases/watcher/node_reorg_check_test.go
Luisfc68, line 1: update the test to validate the reorgs properly, because they should be able to catch the scenario I described in the other comments

Luisfc68 (changes requested):

internal/adapters/dataproviders/bitcoin/mempool_space/api.go
volodymyrzahanych: done

internal/adapters/dataproviders/bitcoin/rpc.go
volodymyrzahanych: test cases added

volodymyrzahanych: test cases added

internal/adapters/entrypoints/watcher/monitoring/metrics.go
volodymyrzahanych: new type introduced 
```
type nodeReorgMetrics struct {
	Depth          *prometheus.GaugeVec
	MaxDepth       *prometheus.GaugeVec
	AboveThreshold *prometheus.GaugeVec
	CheckErrors    *prometheus.CounterVec
	Alerts         *prometheus.CounterVec
}
```

volodymyrzahanych: new type introduced 
```
type nodeReorgMetrics struct {
	Depth          *prometheus.GaugeVec
	MaxDepth       *prometheus.GaugeVec
	AboveThreshold *prometheus.GaugeVec
	CheckErrors    *prometheus.CounterVec
	Alerts         *prometheus.CounterVec
}
```

internal/adapters/entrypoints/watcher/common.go
volodymyrzahanych: I have updated watcher intervals in internal/adapters/entrypoints/watcher/common.go:

```
bitcoinReorgCheckInterval: 1 * time.Minute → 5 * time.Minute
rootstockReorgCheckInterval: 1 * time.Minute → 30 * time.Second
```
5 min for bitcoin allows to reduce load but still detect within a reasonable window.  BTC reorgs are rarer but impactful; when they happen, waiting 10 minutes to notice can be too slow for alerting/mitigation

internal/usecases/watcher/node_reorg_check.go
volodymyrzahanych: done

volodymyrzahanych: done

internal/usecases/watcher/node_reorg_check.go
volodymyrzahanych: In internal/usecases/watcher/node_reorg_check.go I replaced:

```
hw → historyWindow
st → state
hdr → header
```

internal/usecases/watcher/node_reorg_check.go
volodymyrzahanych: In internal/usecases/watcher/node_reorg_check.go:

Removed the useCase.publishReorgCheck(..., 0, ...) call from:
```
initBitcoinState(...)
initRootstockState(...)
```
Moved that publishing to the callers:
```
runBitcoin() now publishes the “0 depth” event after a successful init.
runRootstock() does the same.
```
So the init helpers now only do state initialization (refresh window + set last tip), and the outer flow owns “side effects” like event publishing.

internal/usecases/watcher/node_reorg_check.go
volodymyrzahanych: In internal/usecases/watcher/node_reorg_check.go:

- advanceBitcoinTip(...) now only updates state.lastTipHash/lastTipHeight.
- The window refresh (refreshBitcoinWindow(...)) and the “0 depth” publish are now done in runBitcoin() right after calling advanceBitcoinTip(...).

volodymyrzahanych: In internal/usecases/watcher/node_reorg_check.go:

- advanceBitcoinTip(...) now only updates state.lastTipHash/lastTipHeight.
- The window refresh (refreshBitcoinWindow(...)) and the “0 depth” publish are now done in runBitcoin() right after calling advanceBitcoinTip(...).

internal/usecases/watcher/node_reorg_check.go
volodymyrzahanych: I applied the same change to Rootstock:

- advanceRootstockTip(...) now only updates the tip fields.
- runRootstock() does refreshRootstockWindow(...) + publishes the “0 depth” event.

internal/usecases/watcher/node_reorg_check.go
volodymyrzahanych: In internal/usecases/watcher/node_reorg_check.go I renamed the functions to avoid implying “this is definitely a reorg”:

- bitcoinReorgDepth → bitcoinChainDivergenceDepth
- rootstockReorgDepth → rootstockChainDivergenceDepth

And updated the two call sites:

- reorgDepth := bitcoinChainDivergenceDepth(...)
- reorgDepth := rootstockChainDivergenceDepth(...)

volodymyrzahanych: In internal/usecases/watcher/node_reorg_check.go I renamed the functions to avoid implying “this is definitely a reorg”:

- bitcoinReorgDepth → bitcoinChainDivergenceDepth
- rootstockReorgDepth → rootstockChainDivergenceDepth

And updated the two call sites:

- reorgDepth := bitcoinChainDivergenceDepth(...)
- reorgDepth := rootstockChainDivergenceDepth(...)

internal/adapters/entrypoints/watcher/rootstock_reorg_watcher.go
```diff
@@ -0,0 +1,58 @@
...
+		ticker:             ticker,
+		watcherStopChannel: watcherStopChannel,
+		validationTimeout:  validationTimeout,
+	}
+}
+
+func (watcher *RootstockReorgWatcher) Prepare(ctx context.Context) error { return nil }
+
+func (watcher *RootstockReorgWatcher) Start() {
+watcherLoop:
```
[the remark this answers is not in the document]
volodymyrzahanych: Is the same as in other watchers and fits gofmt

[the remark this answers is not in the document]
volodymyrzahanych: Is the same as in other watchers and fits gofmt

internal/adapters/entrypoints/watcher/monitoring/reorg_metrics_watcher.go
```diff
@@ -0,0 +1,64 @@
...
+}
+
+func (watcher *ReorgMetricsWatcher) Prepare(ctx context.Context) error { return nil }
+
+func (watcher *ReorgMetricsWatcher) Start() {
+	checkCh := watcher.eventBus.Subscribe(blockchain.NodeReorgCheckEventId)
+	errCh := watcher.eventBus.Subscribe(blockchain.NodeReorgCheckErrorEventId)
+	alertCh := watcher.eventBus.Subscribe(blockchain.NodeReorgAlertSentEventId)
+
+metricLoop:
```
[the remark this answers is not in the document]
volodymyrzahanych: Is the same as in other watchers and fits gofmt

internal/usecases/watcher/node_reorg_check.go
```diff
@@ -0,0 +1,434 @@
...
+	reorgDepth uint64,
+	maxDepth uint64,
+) error {
+	if time.Since(useCase.lastAlertByNode[nodeType]) < useCase.alertCooldown {
+		return nil
+	}
+	body := fmt.Sprintf(NodeReorgAlertBodyTemplate, nodeType, reorgDepth, maxDepth)
+	if alertErr := useCase.alertSender.SendAlert(ctx, alerts.AlertSubjectNodeReorg, body, []string{useCase.alertRecipient}); alertErr != nil {
+		log.Errorf("NodeReorgCheckUseCase[%s]: error sending reorg alert: %v", nodeType, alertErr)
+		return useCase.reorgError(nodeType, alertErr)
```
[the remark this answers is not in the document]
volodymyrzahanych: fixed according to suggestions

internal/usecases/watcher/node_reorg_check.go
```diff
@@ -0,0 +1,434 @@
...
+	rpc             blockchain.Rpc
+	alertSender     alerts.AlertSender
+	alertRecipient  string
+	eventBus        entities.EventBus
+	maxDepthByNode  map[entities.NodeType]uint64
+	alertCooldown   time.Duration
+	lastAlertByNode map[entities.NodeType]time.Time
+	bitcoinState    *reorgChainState
+	rootstockState  *reorgChainState
+}
```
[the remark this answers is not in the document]
volodymyrzahanych: [node_reorg_check.go](vscode-webview://04j4imsh71febdiea1enflumt87gm1rr7dds9q9k7bk1rbhqhjiq/internal/usecases/watcher/node_reorg_check.go): Replaced maxDepthByNode map, lastAlertByNode map, bitcoinState, and rootstockState with single maxDepth, lastAlert, and state fields. Constructor now takes one maxDepth instead of two.
[usecase.go](vscode-webview://04j4imsh71febdiea1enflumt87gm1rr7dds9q9k7bk1rbhqhjiq/internal/configuration/registry/usecase.go): Split nodeReorgCheckUseCase into btcReorgCheckUseCase and rskReorgCheckUseCase, each constructed with its own maxDepth.
[watcher.go](vscode-webview://04j4imsh71febdiea1enflumt87gm1rr7dds9q9k7bk1rbhqhjiq/internal/configuration/registry/watcher.go): Wire each watcher to its own use case instance.
[node_reorg_check_test.go](vscode-webview://04j4imsh71febdiea1enflumt87gm1rr7dds9q9k7bk1rbhqhjiq/internal/usecases/watcher/node_reorg_check_test.go): Updated all constructor calls (removed extra arg), and fixed the "alert send failure" test to no longer expect NodeReorgCheckErrorEvent (consistent with the earlier fix).

internal/usecases/watcher/node_reorg_check.go
```diff
@@ -0,0 +1,434 @@
...
+		lowerBound = maxTip - hw + 1
+	}
+
+	for h := minTip; ; {
+		if h < lowerBound {
+			return maxDepth + 1
+		}
+		curHash, err := btc.GetBlockHashAtHeight(int64(h))
+		if err != nil {
+			return maxDepth + 1
```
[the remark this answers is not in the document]
volodymyrzahanych: Already fixed

internal/usecases/watcher/node_reorg_check.go
```diff
@@ -0,0 +1,434 @@
...
+		lowerBound = maxTip - hw + 1
+	}
+
+	for h := minTip; ; {
+		if h < lowerBound {
+			return maxDepth + 1
+		}
+		bi, err := rsk.GetBlockByNumber(ctx, new(big.Int).SetUint64(h))
+		if err != nil {
+			return maxDepth + 1
```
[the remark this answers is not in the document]
volodymyrzahanych: Already fixed

internal/usecases/watcher/node_reorg_check.go
Luisfc68, line 1: A couple things I still see in this file:
- state is shared between chains you should have different states for each one
- last alert is also shared
- since the same use case is being called by two watchers (which is ok) I think you should consider possible concurrent calls to this use case, so you probably need to add the proper mutexes to ensure concurrent calls won't bring any problems

Luisfc68 (changes requested):

internal/usecases/watcher/node_reorg_check.go
volodymyrzahanych: done

internal/usecases/watcher/node_reorg_check.go
```diff
@@ -87,15 +92,22 @@ func (useCase *NodeReorgCheckUseCase) handleReorgAlert(
 	reorgDepth uint64,
 	maxDepth uint64,
 ) error {
-	if time.Since(useCase.lastAlert) < useCase.alertCooldown {
+	var lastAlert *time.Time
```
Luisfc68: maybe you don't need to declare a pointer here, look the usage below, you always end up using the value instead of the ref

Luisfc68 (changes requested):
I think maybe after fixing the latest comment the code will be ready

internal/usecases/watcher/node_reorg_check.go
volodymyrzahanych: done

volodymyrzahanych: done

Luisfc68 (approved):

PR #950, opened 2026-04-10
fix fly2248 update lps libraries to latest version except for mongodb

AndresQuijano:
## What

Upgrades all Go module dependencies to their latest stable versions, excluding `mongo-driver` (intentionally pinned at v1.17.4).

### Go directive: 1.23.8 → 1.25.9

The `go` directive in `go.mod` was raised automatically by `go mod tidy` after upgrading `github.com/go-playground/validator/v10` to **v10.30.2**. That module's own `go.mod` declares `go 1.25.0`, and Go's toolchain requires the consuming module's directive to be ≥ all its dependencies'. For reference, the AWS SDK v2 upgrade alone would have only required Go 1.24 — it was specifically `validator/v10@v10.30.2` that pushed the floor to 1.25.
After running govulncheck, 21 vulnerabilities were found so it was updated again to 1.25.9.

### Direct dependency upgrades

| Module | Old | New |
|---|---|---|
| `aws/aws-sdk-go-v2` (+ all sub-packages) | v1.21.2 | v1.41.5 |
| `aws/smithy-go` | v1.15.0 | v1.24.3 |
| `ethereum/go-ethereum` | v1.14.13 | v1.17.2 |
| `gorilla/csrf` | v1.7.2 | v1.7.2 (no changes because new one has vulnerabilities) |
| `awnumar/memguard` | v0.22.5 | v0.23.0 |
| `btcsuite/btcd` | v0.24.2 | v0.25.0 |
| `go-playground/validator/v10` | v10.26.0 | v10.30.2 |
| `prometheus/client_golang` | v1.22.0 | v1.23.2 |
| `sirupsen/logrus` | v1.9.3 | v1.9.4 |
| `golang.org/x/crypto` | v0.40.0 | v0.49.0 |
| `golang.org/x/sync` | v0.16.0 | v0.20.0 |
| `golang.org/x/term` | v0.33.0 | v0.41.0 |

### Code changes

**AWS SDK v2 endpoint resolver migration** (`aws.go`, `secrets/aws.go`):
The deprecated `aws.EndpointResolverWithOptionsFunc` + `config.WithEndpointResolverWithOptions` pattern was replaced with the per-service `BaseEndpoint` option on `secretsmanager.Options`. `GetAwsConfig` no longer takes an `Environment` parameter; a new `AwsLocalEndpoint` helper returns the LocalStack endpoint for regtest mode.

**abigen tool pin** (`Makefile`):
Updated the `abigen` install pin to the commit hash for go-ethereum v1.17.2. Existing contract bindings compile cleanly without regeneration.

**Docker builder images** (`docker-compose/lps/Dockerfile`, `docker-compose/utils/Dockerfile`):
Updated the Go builder base image from `golang:1.23.8` to `golang:1.25.0` to match the `go.mod` directive and prevent build failures inside Docker where `GOTOOLCHAIN=local` is the default.

**RSKj version** (`docker-compose/rskj/Dockerfile`):
Updated RSKj from `8.1.0` to `9.0.1`.

## Update in tests

Migrated all watcher test files from `assert.Eventually` to `assert.EventuallyWithT` to comply with the breaking behavior change introduced in `testify v1.11.0`, which now evaluates the condition immediately at `t=0` instead of waiting for the first tick.

### Root cause

`testify v1.11.0` changed `assert.Eventually` to check the condition at `t=0`. The existing tests passed mock assertions (e.g. `mock.AssertExpectations(t)`) and log checks directly inside the callback. When the condition was false at `t=0`, these called `t.Errorf` on the real `*testing.T`, permanently marking the test as failed even though `Eventually` would have retried and eventually succeeded.

### `test/utils.go` — `AssertLogContains`

- Removed the `t.Errorf("No log message found")` call when the log buffer is empty. The closure now returns `false` silently, allowing `EventuallyWithT` to retry on the next tick.
- Replaced `require.NoError(t, err, ...)` with a plain `if err != nil { return false }`. `require.NoError` calls `t.FailNow()` which invokes `runtime.Goexit()` — this is unsafe when called from a non-test goroutine (such as the one spawned internally by `Eventually`).

### `internal/adapters/entrypoints/watcher/common_test.go` — `mockCollectT` wrapper

Added a `mockCollectT` wrapper struct that embeds `*assert.CollectT` and adds a no-op `Logf` method. This bridges the gap between `assert.CollectT` (provided by `EventuallyWithT`) and `mock.TestingT` (required by `mock.AssertExpectations`, `mock.AssertNotCalled`), which additionally requires `Logf`.

### Test files updated

The following 9 test files were migrated:

- `eclipse_test.go`
- `btc_release_watcher_test.go`
- `quote_cleaner_test.go`
- `penalization_alert_test.go`
- `pegout_bridge_watcher_test.go`
- `pegout_btc_watcher_test.go`
- `pegout_rsk_watcher_test.go`
- `pegin_bridge_watcher_test.go`
- `pegin_btc_deposit_watcher_test.go`

**Migration pattern applied consistently across all files:**

- Mock expectations (`AssertExpectations`, `AssertNotCalled`) and state assertions (`assert.Equal`, `assert.True`, `assert.False`, `assert.Empty`) are placed **inside** the `EventuallyWithT` callback using `collect` / `mockCollectT{collect}` as the `TestingT`. Errors are accumulated per-tick and do not permanently fail the test during retries.
- Log assertions (`test.AssertLogContains`) are called **after** `EventuallyWithT` returns. This is safe because log writes in all watchers are synchronous — they happen in the same goroutine as the mock calls, so once mock expectations are confirmed met, all log output is already in the buffer.
- For cases where the log check is the **only** condition (no mock to wait for), `assert.Eventually` is retained. With the `AssertLogContains` fix in place, an empty buffer at `t=0` now returns `false` silently instead of permanently failing.

## Why
Reduce some of the tech debt

## Type of Change
- [x] Breaking change 
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test updates
- [ ] Security fix

- [x] Management UI / API
- [ ] PegIn flow
- [ ] PegOut flow
- [ ] Utility scripts
- [x] Configuration files
- [ ] Metrics and alerting

## Related Issues
[Jira ticket](https://rsklabs.atlassian.net/browse/FLY-2248)

## Test
Probably the full app should be tested to ensure everything is on place


test/utils.go
```diff
@@ -127,11 +127,12 @@ func AssertLogContains(t *testing.T, expected string) (assertFunc func() bool) {
 	log.SetOutput(buff)
 	return func() bool {
 		if buff.Len() == 0 {
-			t.Errorf("No log message found")
 			return false
 		}
 		_, err := buff.Read(message)
-		require.NoError(t, err, "Error reading log message")
+		if err != nil {
```
Luisfc68, line 133: any specific reason for this change? wouldn't `require.NoError` show a better trace if the read fails?

Luisfc68 (approved):

test/utils.go
AndresQuijano: `require.NoError` calls `t.FailNow()` under the hood. `.Eventually` runs the inner function of `AssertLogContains` in a separated goroutine so `.FailNow()` will execute` .GoExit() `and will never return the bool to` Eventually`. Returning true/false is the standard recommendation. 
Note that this bug was already there before the migration but was silently living for free because `buff.Read` (line 132) never errors.

PR #953, opened 2026-04-10
Fix refund regtest

volodymyrzahanych:
## What
Panic in refund pegout flow

## Why

## Type of Change
- [ ] Breaking change 
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test updates
- [ ] Security fix

- [ ] Management UI / API
- [ ] PegIn flow
- [x] PegOut flow
- [ ] Utility scripts
- [ ] Configuration files
- [ ] Metrics and alerting

## Related Issues
[Jira ticket](https://rsklabs.atlassian.net/browse/FLY-2266)

## How to test
Check that LPS does not crash when  the RSK node returns an rpc.DataError whose ErrorData() 
 is "0x" (no payload, not even a 4-byte selector) 

## Reviewer Guidelines

## Additional Notes

### DIFF
--- a/internal/adapters/dataproviders/rootstock/common.go
+++ b/internal/adapters/dataproviders/rootstock/common.go
@@ -23,8 +23,11 @@ const (
 )
 
 // ErrShortRevertData is returned by ParseRevertReason when the revert payload
-// is shorter than a 4-byte ABI selector. This typically indicates a transient
-// bridge state rather than a named contract error.
+// is shorter than a 4-byte ABI selector, so the error cannot be decoded into a
+// named custom error (or revert string). This can happen for multiple reasons
+// (e.g. ABI decode failures, calling an address with no code, or other opaque
+// reverts), and should not be interpreted as a specific on-chain condition by
+// the parser.
 var ErrShortRevertData = errors.New("revert data shorter than ABI selector")
 
 var DefaultRetryParams = RetryParams{
--- a/internal/adapters/dataproviders/rootstock/common.go
+++ b/internal/adapters/dataproviders/rootstock/common.go
@@ -22,6 +22,11 @@ const (
 	rpcCallRetrySleep = 1 * time.Minute
 )
 
+// ErrShortRevertData is returned by ParseRevertReason when the revert payload
+// is shorter than a 4-byte ABI selector. This typically indicates a transient
+// bridge state rather than a named contract error.
+var ErrShortRevertData = errors.New("revert data shorter than ABI selector")
+
 var DefaultRetryParams = RetryParams{
 	Retries: rpcCallRetryMax,
 	Sleep:   rpcCallRetrySleep,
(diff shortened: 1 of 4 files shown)


internal/adapters/dataproviders/rootstock/common.go
```diff
@@ -22,6 +22,11 @@ const (
 	rpcCallRetrySleep = 1 * time.Minute
 )
 
+// ErrShortRevertData is returned by ParseRevertReason when the revert payload
```
Luisfc68, line 25: Can you elaborate on this line?
`This typically indicates a transient bridge state rather than a named contract error.`
What is the rationale to consider the empty payload is related to a transient bridge state? The empty payload seems to be generated for other reasons not related to the bridge state

internal/adapters/dataproviders/rootstock/pegout_contract.go
```diff
@@ -214,7 +214,10 @@ func (pegoutContract *pegoutContractImpl) RefundPegout(txConfig blockchain.Trans
 	}
 	_, revert := pegoutContract.contract.CallRaw(&bind.CallOpts{From: pegoutContract.signer.Address()}, callData)
 	parsedRevert, err := ParseRevertReason(pegoutContract.abis.PegOut, revert)
-	if err != nil && parsedRevert == nil {
+	if errors.Is(err, ErrShortRevertData) {
```
Luisfc68: Is there any specific reason why you treat this as an `WaitingForBridgeError`? The `WaitingForBridgeError` is reserved for the scenario where the bridge hasn't been updated with the latest blocks. Ideally, we should only return it on known scenarios, an empty payload is not a known scenario

Luisfc68 (changes requested):
I think this fix should be targeted to v2.5.1 instead of v2.6.0 as the issue is present in the current version

internal/adapters/dataproviders/rootstock/common.go
volodymyrzahanych: Comment was updated

internal/adapters/dataproviders/rootstock/pegout_contract.go
volodymyrzahanych: ErrShortRevertData was seperated

internal/adapters/dataproviders/rootstock/pegout_contract.go
```diff
@@ -191,7 +191,10 @@ func (pegoutContract *pegoutContractImpl) RefundPegout(txConfig blockchain.Trans
 		params.MerkleBranchHashes,
 	)
 	parsedRevert, err := ParseRevertReason(pegoutContract.abis.PegOut, revert)
-	if err != nil && parsedRevert == nil {
+	if errors.Is(err, ErrShortRevertData) {
```
Luisfc68: which is the rationale to add a separate clause for this error? If is just for the custom message I think is not necessary at all, it only adds noise which may affect posterior debugging, whatever function is consuming this already has enough information if you wrap the error, and outer layers are supposed to log errors so we can construct the whole stack trace, I don't think this line give any valuable information by itself that can be used while debugging, actually, if you see the error as part of the chain because of a log in a layer above you'll easily determine it was produced here by tracing the call stack

Luisfc68: If you still want to maintain the log, then I recommend to at least reduce its level to trace

Luisfc68 (changes requested):

internal/adapters/dataproviders/rootstock/pegout_contract.go
volodymyrzahanych: My idea was to mark that special case in logs as earlier as possible  since that took so much time to find the exact  reason for 0x empty revert. But I agree that additional log might be a little help as it gives no additional valuable data for investigation

Luisfc68 (approved):

PR #958, opened 2026-04-15
fix(FLY-2101): fix misleading /pegin/recommended validation error

volodymyrzahanych:
## What
Unclear error message and misleading range validation in /pegin/recommended endpoint

## Why
When executing the /pegin/recommended endpoint with an amount parameter equal to the lower bound of the valid range (6000000000000000), the API returns a 400 Bad Request with the following message:
{
  "message": "RecommendedPegin: recommended amount 5999310330477010 is out of range: amount out of range [6000000000000000, 1000000000000000000]",
  "details": {},
  "timestamp": 1762793364,
  "recoverable": true
}
This message is confusing because the user provides a valid amount (equal to the minimum allowed), but the API still rejects it due to internal fee deductions that reduce the effective amount below the minimum threshold.
As a result, users are unable to determine what amount would actually be accepted, and the current message does not clearly explain why the request fails.

## Type of Change
- [ ] Breaking change 
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test updates
- [ ] Security fix

- [x] Management UI / API
- [ ] PegIn flow
- [ ] PegOut flow
- [ ] Utility scripts
- [ ] Configuration files
- [ ] Metrics and alerting

## Related Issues
[Jira ticket](https://rsklabs.atlassian.net/browse/FLY-2101)

## How to test
Run corresponding unit  tests.
Use regtest env for:
### 1. Happy path - 1 RBTC, clearly within range
echo "=== 1. Happy path (1 RBTC) ==="
curl -s 'http://localhost:8080/pegin/recommended?amount=1000000000000000000' | jq .

### 2. Below MinValue (0.1 RBTC < 0.6 RBTC min) → AmountOutOfRangeError
echo "=== 2. Below MinValue (0.1 RBTC) ==="
curl -s 'http://localhost:8080/pegin/recommended?amount=100000000000000000' | jq .

### 3. Exactly at MinValue (0.6 RBTC) → after fees result < MinValue → EffectiveAmountTooLowError
echo "=== 3. At MinValue (0.6 RBTC) ==="
curl -s 'http://localhost:8080/pegin/recommended?amount=600000000000000000' | jq .

### 4. Above MaxValue (20 RBTC > 10 RBTC max) → AmountOutOfRangeError
echo "=== 4. Above MaxValue (20 RBTC) ==="
curl -s 'http://localhost:8080/pegin/recommended?amount=20000000000000000000' | jq .
2>&1

## Reviewer Guidelines

## Additional Notes

### DIFF
--- a/docker-compose/local/docker-compose.yml
+++ b/docker-compose/local/docker-compose.yml
@@ -55,7 +55,7 @@ services:
       context: ../rskj
       args:
         UID: "${LPS_UID}"
-        RSKJ_VERSION: "${RSKJ_VERSION}"
+        RSKJ_VERSION: "${RSKJ_VERSION:-8.1.0~noble}"
     pull_policy: build
     image: rskj:latest
     container_name: rskj01
--- a/docker-compose/local/docker-compose.yml
+++ b/docker-compose/local/docker-compose.yml
@@ -55,6 +55,7 @@ services:
       context: ../rskj
       args:
         UID: "${LPS_UID}"
+        RSKJ_VERSION: "${RSKJ_VERSION}"
     pull_policy: build
     image: rskj:latest
     container_name: rskj01
--- a/internal/usecases/pegin/recommended_pegin.go
+++ b/internal/usecases/pegin/recommended_pegin.go
@@ -74,15 +74,15 @@ func (useCase *RecommendedPeginUseCase) Run(
 	// It is an estimate because gas is re-estimated against the original user amount, not the
 	// suggested one, so the follow-up request may yield a slightly different fee.
 	// requiredNet = ceil(MinValue * totalPercentages / scale)
-	// suggestedAmount = gasFee + fixedFee + requiredNet
+	// minimumAcceptable = gasFee + fixedFee + requiredNet
 	minValue := config.MinValue.AsBigInt()
 	requiredNet := new(big.Int).Mul(minValue, totalPercentages)
 	requiredNet.Add(requiredNet, new(big.Int).Sub(big.NewInt(useCase.scale), big.NewInt(1)))
 	requiredNet.Quo(requiredNet, big.NewInt(useCase.scale))
-	suggestedAmount := new(big.Int).Add(gasFeeEstimation, fixedCallFeeEstimation)
-	suggestedAmount.Add(suggestedAmount, requiredNet)
+	minimumAcceptable := new(big.Int).Add(gasFeeEstimation, fixedCallFeeEstimation)
+	minimumAcceptable.Add(minimumAcceptable, requiredNet)
 
-	if err = useCase.validateRecommendedValue(ctx, config, result, suggestedAmount); err != nil {
+	if err = useCase.validateRecommendedValue(ctx, config, result, minimumAcceptable); err != nil {
 		return usecases.RecommendedOperationResult{}, err
 	}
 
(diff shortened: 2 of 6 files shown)


internal/usecases/pegin/recommended_pegin.go
```diff
@@ -128,21 +140,26 @@ func (useCase *RecommendedPeginUseCase) validateRecommendedValue(
...
 	result *big.Int,
+	suggestedAmount *big.Int,
 ) error {
-	var err error
-
-	if err = config.ValidateAmount(entities.NewBigWei(result)); err != nil {
-		err = fmt.Errorf("recommended amount %s is out of range: %w", entities.NewBigWei(result).String(), err)
-		return usecases.WrapUseCaseError(usecases.RecommendedPeginId, err)
+	minValue := config.MinValue.AsBigInt()
+	if result.Cmp(minValue) < 0 {
```
Luisfc68: this block can be rewritten to avoid the nested if by using if (result is smaller than min and suggested is higher than max)/ else if (result is smaller than min)

internal/usecases/pegin/recommended_pegin.go
Luisfc68, line 1: The name `suggestedAmount` can be confusing for the context of this use case, the use case is already to provide a suggestion based on the input and `suggestedAmount` is only used if the suggestion doesn't meet the criteria. I'd recommend to change it by minimumAcceptable or something like that to reflect is **not** the same as the suggestion the use case intents to do

Luisfc68 (changes requested):

internal/usecases/pegin/recommended_pegin_test.go
```diff
@@ -92,14 +94,32 @@ func TestRecommendedPeginUseCase_Run(t *testing.T) {
...
+		// and the suggested amount exceeds MaxValue, so no valid input exists in the provider range.
+		result, err = useCase.Run(context.Background(), modifiedConfig.MinValue, test.AnyRskAddress, data)
 		require.ErrorIs(t, err, liquidity_provider.AmountOutOfRangeError)
 		assert.Empty(t, result)
+		modifiedLimitLp.AssertNotCalled(t, "HasPeginLiquidity")
+	})
+	t.Run("should return EffectiveAmountTooLowError with exact suggested amount when userBalance equals MinValue", func(t *testing.T) {
+		// MinValue=1000, gasFee=20000*100=2000000, fixedFee=100, feePercentage=1.25%, scale=10000
+		// scaledCallFeePercentage=125, totalPercentages=10125
+		// result = floor((1000-2000000-100)*10000/10125) = -1974419
```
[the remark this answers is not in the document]
volodymyrzahanych: fixed

internal/usecases/pegin/recommended_pegin.go
```diff
@@ -128,21 +140,26 @@ func (useCase *RecommendedPeginUseCase) validateRecommendedValue(
...
+		if suggestedAmount.Cmp(config.MaxValue.AsBigInt()) > 0 {
+			return usecases.WrapUseCaseError(usecases.RecommendedPeginId,
+				fmt.Errorf("no valid input amount exists within provider range: %w", liquidity_provider.AmountOutOfRangeError),
```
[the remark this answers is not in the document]
volodymyrzahanych: fixed

docker-compose/local/docker-compose.yml
```diff
@@ -55,6 +55,7 @@ services:
       context: ../rskj
       args:
         UID: "${LPS_UID}"
+        RSKJ_VERSION: "${RSKJ_VERSION}"
```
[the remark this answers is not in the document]
volodymyrzahanych: fixed

internal/usecases/pegin/recommended_pegin.go
volodymyrzahanych: fixed according to the suggestions

internal/usecases/pegin/recommended_pegin.go
volodymyrzahanych: renamed according to the suggestions

Luisfc68 (approved):

PR #963, opened 2026-05-01
fix(FLY-2269): redact secrets in environment string output

volodymyrzahanych:
## What
Currently the LPS logs the environment in debug mode during the startup. Since the environment includes some keys they are shown in the log as well.

## Why
The purpose of this task is to create (or modify) the String function of the environment structure to censor those keys.

## Type of Change
- [ ] Breaking change 
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test updates
- [ ] Security fix

- [ ] Management UI / API
- [ ] PegIn flow
- [ ] PegOut flow
- [ ] Utility scripts
- [x] Configuration files
- [ ] Metrics and alerting

## Related Issues
[Jira ticket](https://rsklabs.atlassian.net/browse/FLY-2269)

## How to test
- Run unit tests (including internal/configuration/environment tests).
- Start LPS and review bootstrap logs to confirm secret env values are masked and non-secret values still print correctly.

## Reviewer Guidelines

## Additional Notes


Luisfc68 (changes requested):
Why is the implementation of this so complex? using a tag and reflection on that tag is too much and therefore too error prone for the issue we're trying to solve. Creating a custom String() function that manually censors the fields we want to censor is enough, there is no need to use reflection for this problem in particular

Luisfc68:
For example this is the same outcome without the need of introducing code that is expensive to maintain

image

volodymyrzahanych:
Actually I considered the simple solution from the beginning but decided to suggest more complex one with tags for review. There are  trade-offs as always. Simple solution is more explicit but needs to maintain a list of secret fields. Tag solution with reflection is more automatic but more complex.   I rewrote  to the simple one.

Luisfc68:
There is no need to maintain a explicit list. This can be addressed cleanly by properly using the features of the language. Here is a extended version of the example I provided in my previous comment 
image

volodymyrzahanych:
A solution with the sensitive type is a quite elegant one I agree.   But it is not so simple as it seems.  It might need more code refactoring since, for example, SensitiveInt is not int. Any arithmetic, comparison with literals, or passing to a function expecting int requires an explicit cast. On the other hand a list solution is ~25 lines of straightforward Go, no reflection, no struct tags, no domain-layer churn. The Sensitive-type proposal would touch dozens of files for a narrower benefit and it does not solve all possible sensitive key cases as it protects only one  leak vector .  Looks like this simple one place log fix brings us to some architectural decisions we need to take: should we develop a comprehensive sensitive key protection in lps including json.Marshal, MongoDB BSON marshaling, etc or just fix existing log entry where secrets leak right now?

Luisfc68:
Both the list and the Sensitive type solutions protect only the log leak vector. The difference is that the Sensitive type is more extensible to protect us from other cases in the future, in that regard I think the price of adding the casting is fair (also because that would only affect the configuration layer as layers below shouldn't interact with the environment package). Anyway, maintaining the explicit list also solves the issue and is not unnecessarily complex as the first approach, so I don't have any problem with accepting it even though don't I think is the best solution

volodymyrzahanych:
1. Due to update to wrong base branch this pr is closed and the new one created  [PR984 ](https://github.com/rsksmart/liquidity-provider-server/pull/984)

PR #697, opened 2025-05-13
feat: add functionality to accept pegin quote without CAPTCHA

AndresQuijano:
## What
Add an endpoint and full flow to accept a pegin quote without CAPTCHA by introducing an additional optional parameter to the existing use case. This parameter is a signature from the owner account.

## Why
This change is part of the functionality to automate pegin and pegout flows

More info [here](https://www.notion.so/iovlabs/Flyover-Programmatic-PegIn-and-PegOut-capability-implementation-proposal-1b2c132873f980bfad2cf6c79dfbdb8b)

[Jira tk](https://rsklabs.atlassian.net/browse/GBI-2613)

PR #964, opened 2026-05-05
FLY-2267 Flyover SDK - Signature verification fails when trusted address is saved with checksum

TravellerOnTheRun:
## What
Trusted-account flows now treat EVM addresses in a single canonical form (`0x` + 40 lowercase hex) end-to-end:

- **`NormalizeEthereumAddress`** (`internal/entities/blockchain/`) validates with go-ethereum `IsHexAddress` and returns the lowercase canonical string.
- **`RecoverSignerAddress`** returns the same lowercase form as EIP-55 `Hex()` so recovered signers match DB lookups.
- **Trusted-account use cases** (add / update / delete / get) normalize the address before signature verification and repository calls.
- **Mongo** `trustedAccounts` reads use `$expr` with `$toLower` on the stored `address` field so existing documents saved with mixed or EIP-55 casing still match.
- **`InvalidTrustedAccountAddressError`** for invalid hex; **management handlers** map it to **HTTP 400** with details.
- **Unit tests** updated for normalization, Mongo filter shape, handlers, and related peg flows that mock `GetTrustedAccount`.

## Why
Mongo previously matched `address` exactly. Rows stored with different string casing than the address produced by **`RecoverSignerAddress`** (mixed-case EIP-55) caused lookups to miss and failures that looked like signature/tamper issues.

## Type of Change
- [ ] Breaking change 
- [ ] Documentation update
- [ ] Performance improvement
- [x] Test updates
- [ ] Security fix

- [x] Management UI / API
- [x] PegIn flow
- [x] PegOut flow
- [ ] Utility scripts
- [ ] Configuration files
- [ ] Metrics and alerting

## Related Issues
None linked from this PR description (link a Jira ticket here if applicable).

## How to test
- Run unit tests (excludes integration suites that need local config):
  - `go test ./internal/... ./pkg/...`
- **Management:** add/update/delete/get trusted account with the same EOA expressed as checksummed (EIP-55) vs all-lowercase; both should resolve the same document and signatures should succeed when the payload matches the normalized key.
- **Mongo:** optional manual check that an existing document with mixed-case `address` is returned by get/delete when querying with a different casing of the same address.

## Reviewer Guidelines
- Confirm **`$expr` + `$toLower`** on `trustedAccounts` is acceptable for your indexes/workload (may differ from a simple equality filter on `address`).
- Confirm **400** for invalid address strings is the desired API contract vs other error codes.
- Peg paths that call `GetTrustedAccount` now expect/use normalized addresses in mocks—sanity-check no behavioral drift beyond casing.

N/A (server-side / API change).

## Additional Notes
- **Base branch:** `v2.5.1` (release line this fix targets).
- **Commit:** conventional `fix(trusted-accounts): ...` on branch `fly-2267`.


internal/entities/blockchain/normalize_address.go
```diff
@@ -0,0 +1,18 @@
+package blockchain
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/ethereum/go-ethereum/common"
```
volodymyrzahanych: This is the first production import of any `go-ethereum/*` package under `internal/entities/`. Per the entity-layer-boundaries rule (PR #697: *"the entity layer must not import infrastructure packages"*), this isn't allowed. It's also unnecessary — `blockchain.IsRskAddress` already exists in this package at [rootstock.go:42](internal/entities/blockchain/rootstock.go#L42) and is used by 8+ call sites. Suggest dropping this file entirely and adding the helper to `rootstock.go` next to `IsRskAddress`:

 ```go
// internal/entities/blockchain/rootstock.go
 func NormalizeRskAddress(addr string) (string, error) {
    trimmed := strings.TrimSpace(addr)
    if !IsRskAddress(trimmed) {
         return "", fmt.Errorf("%w: %q", InvalidAddressError, addr)
    }
    return strings.ToLower(trimmed), nil
}
 ```

All current callers pass `0x` + 40 hex (admin API request + `RecoverSignerAddress` output), so the behavior is identical without `common.IsHexAddress` / `common.HexToAddress(...).Hex()`.

internal/entities/liquidity_provider/trusted_account.go
```diff
@@ -8,9 +8,10 @@ import (
 )
 
 var (
-	TrustedAccountNotFoundError  = errors.New("trusted account not found")
-	DuplicateTrustedAccountError = errors.New("trusted account already exists")
-	TamperedTrustedAccountError  = errors.New("trusted account signature verification failed")
+	TrustedAccountNotFoundError        = errors.New("trusted account not found")
+	DuplicateTrustedAccountError       = errors.New("trusted account already exists")
+	TamperedTrustedAccountError        = errors.New("trusted account signature verification failed")
+	InvalidTrustedAccountAddressError  = errors.New("invalid trusted account address")
```
volodymyrzahanych, line 14: The go-patterns anti-patterns list says: *"Don't define a new sentinel error if one already exists."* `blockchain.InvalidAddressError` already covers "invalid address" and is mapped at [rootstock/common.go:82](internal/adapters/dataproviders/rootstock/common.go#L82) and [discovery_contract.go:78](internal/adapters/dataproviders/rootstock/discovery_contract.go#L78). Two options:

 - **Drop `InvalidTrustedAccountAddressError` entirely** and have handlers match on `blockchain.InvalidAddressError` directly while still rendering a trusted-account-flavored message string.
- **Keep the domain-flavored name but wrap the base error** so `errors.Is(err, blockchain.InvalidAddressError)` continues to hold and the chain stays inspectable end-to-end:

 ```go
var InvalidTrustedAccountAddressError = fmt.Errorf("%w: trusted account address", blockchain.InvalidAddressError)
 ```

internal/usecases/liquidity_provider/add_trusted_account.go
```diff
@@ -27,6 +29,12 @@ func NewAddTrustedAccountUseCase(
 }
 
 func (useCase *AddTrustedAccountUseCase) Run(ctx context.Context, account liquidity_provider.TrustedAccountDetails) error {
+	normalized, err := blockchain.NormalizeEthereumAddress(account.Address)
+	if err != nil {
+		return usecases.WrapUseCaseError(usecases.AddTrustedAccountId,
+			errors.Join(liquidity_provider.InvalidTrustedAccountAddressError, err))
+	}
```
volodymyrzahanych: Identical 4-line block here and in the other three trusted-account use cases (`Add`, `Update`, `Delete`, `Get`). Matches the shape of existing shared helpers in [`usecases/common.go`](internal/usecases/common.go) — `ValidateMinLockValue`, `ValidatePositiveWeiValues`, `CheckPauseState`. Extract:

```go
 // usecases/common.go
func NormalizeTrustedAccountAddress(useCase UseCaseId, addr string) (string, error) {
     normalized, err := blockchain.NormalizeRskAddress(addr)
     if err != nil {
         return "", WrapUseCaseError(useCase, errors.Join(liquidity_provider.InvalidTrustedAccountAddressError, err))
     }
     return normalized, nil
 }
 ```

Each use case becomes a one-liner. Note: `GetTrustedAccountUseCase.Run` currently returns this error without `WrapUseCaseError` — that's a pre-existing inconsistency in `get_trusted_account.go`, not introduced by this PR, so feel free to address separately if you want consistency.

File: internal/usecases/liquidity_provider/update_trusted_account.go [hunk 4]
```diff
@@ -27,6 +29,12 @@ func NewUpdateTrustedAccountUseCase(
 }
 
 func (useCase *UpdateTrustedAccountUseCase) Run(ctx context.Context, account liquidity_provider.TrustedAccountDetails) error {
+	normalized, err := blockchain.NormalizeEthereumAddress(account.Address)
+	if err != nil {
+		return usecases.WrapUseCaseError(usecases.UpdateTrustedAccountId,
+			errors.Join(liquidity_provider.InvalidTrustedAccountAddressError, err))
+	}
```
volodymyrzahanych: the same recommendations as in add_trusted_account.go

internal/usecases/liquidity_provider/delete_trusted_account.go
```diff
@@ -20,7 +22,12 @@ func NewDeleteTrustedAccountUseCase(
 }
 
 func (useCase *DeleteTrustedAccountUseCase) Run(ctx context.Context, address string) error {
-	err := useCase.trustedAccountRepository.DeleteTrustedAccount(ctx, address)
+	normalized, err := blockchain.NormalizeEthereumAddress(address)
+	if err != nil {
+		return usecases.WrapUseCaseError(usecases.DeleteTrustedAccountId,
+			errors.Join(liquidity_provider.InvalidTrustedAccountAddressError, err))
+	}
```
volodymyrzahanych: the same recommendations as in add_trusted_account.go

internal/usecases/liquidity_provider/get_trusted_account.go
```diff
@@ -27,10 +28,14 @@ func NewGetTrustedAccountUseCase(
 }
 
 func (useCase *GetTrustedAccountUseCase) Run(ctx context.Context, address string) (*entities.Signed[liquidity_provider.TrustedAccountDetails], error) {
+	normalized, err := blockchain.NormalizeEthereumAddress(address)
+	if err != nil {
+		return nil, errors.Join(liquidity_provider.InvalidTrustedAccountAddressError, err)
+	}
```
volodymyrzahanych: the same recommendations as in add_trusted_account.go

volodymyrzahanych:
Even though the unit tests cover the lookup, the recover-then-store invariant, and the EIP-55 case, regtest is where we actually exercise the whole signed loop with a real wallet: a script signs the EIP-712 quote with the user's key, RecoverSignerAddress lowercases that address, Mongo answers the $toLower $expr query against an admin-inserted mixed-case row, ValidateConfiguration re-hashes the stored payload and validates the LP's signature, the retained quote gets written with the now-canonical owner_account_address, and the BTC/RBTC state machine advances through the watchers. That chain crosses go-ethereum signing, the LBC EIP-712 hashing on RSKj, MongoDB collation behavior, and the locking-cap query — none of which the unit tests touch — and it's exactly the chain the reporter's "trusted account signature verification failed" bug originated from. Running it end-to-end is the only way to prove the symptom is gone for the specific path the user reported, catch any subtle interaction with the still-case-sensitive GetRetainedQuotesForAddress query, and confirm no regression in the happy-path pegin/pegout flow that ships in the same PR.

TravellerOnTheRun:
> Even though the unit tests cover the lookup, the recover-then-store invariant, and the EIP-55 case, regtest is where we actually exercise the whole signed loop with a real wallet: a script signs the EIP-712 quote with the user's key, RecoverSignerAddress lowercases that address, Mongo answers the $toLower $expr query against an admin-inserted mixed-case row, ValidateConfiguration re-hashes the stored payload and validates the LP's signature, the retained quote gets written with the now-canonical owner_account_address, and the BTC/RBTC state machine advances through the watchers. That chain crosses go-ethereum signing, the LBC EIP-712 hashing on RSKj, MongoDB collation behavior, and the locking-cap query — none of which the unit tests touch — and it's exactly the chain the reporter's "trusted account signature verification failed" bug originated from. Running it end-to-end is the only way to prove the symptom is gone for the specific path the user reported, catch any subtle interaction with the still-case-sensitive GetRetainedQuotesForAddress query, and confirm no regression in the happy-path pegin/pegout flow that ships in the same PR.

Thanks for the review — I addressed the feedback below.

---

### What was wrong

Authenticated accept could fail with `trusted account signature verification failed` even when the EIP-712 signature was valid.

The signature was fine. The failure was a **case mismatch on lookup**:

| Step | Address form |
|------|----------------|
| Admin registers trusted account (often EIP-55 / mixed case) | e.g. `0x0C48e53E5852780Cb0a5Bf168eb92DE3B88Ebe89` |
| `RecoverSignerAddress` after signing | lowercase, e.g. `0x0c48e53e5852780cb0a5bf168eb92de3b88ebe89` |
| Mongo lookup (before fix) | exact string match → **miss** → `TamperedTrustedAccountError` |

---

### What I did (and why)

**1. Removed `normalize_address.go` (go-ethereum in entities)**  
Per feedback: keep `entities/blockchain` free of go-ethereum and avoid layering/import-cycle issues. I deleted `NormalizeEthereumAddress` and its tests.

**2. Added `NormalizeRskAddress` in `rootstock.go` (next to `IsRskAddress`)**  
Single canonical normalizer for RSK account addresses: validate with `IsRskAddress`, return lowercase `0x` + 40 hex, wrap invalid input in `InvalidAddressError`. Tests live in `rootstock_test.go` as `TestNormalizeRskAddress`.

**3. Added `NormalizeTrustedAccountAddress` in `usecases/common.go`**  
Shared helper for trusted-account CRUD so add/update/delete/get don’t duplicate logic. On failure: `errors.Join(InvalidTrustedAccountAddressError, blockchain.InvalidAddressError, err)` so callers can `errors.Is` either sentinel.

**4. Mongo `trustedAccountAddressFilter`**  
Lookup key is normalized with `NormalizeRskAddress`; filter uses `$toLower` on stored `address` so rows registered with any casing still match the recovered signer.

**5. Error joining at use-case / mongo layers (not in `entities/blockchain`)**  
`InvalidTrustedAccountAddressError` cannot wrap `InvalidAddressError` inside the entity package without an import cycle (`blockchain` ↔ `liquidity_provider` via `lbc.go`). I join both errors where addresses are handled to preserve the intended `errors.Is` behavior.

**Accept pegin/pegout flows**  
No change required for the reported bug: `RecoverSignerAddress` still returns lowercase; **`GetTrustedAccount` → mongo filter** is now case-insensitive, which fixes authenticated accept.

---

### Optional follow-up (not in this PR): set `trustedAccount.Address` to the recovered signer

I considered (and left out) overwriting the loaded trusted-account address with the recovered signer address in `getTrustedAccount`, e.g. after a successful lookup:

```go
trustedAccount, err := ...GetTrustedAccount(ctx, address)
// ...
trustedAccount.Address = address // recovered lowercase from signature
return trustedAccount, nil
```

**Why it could be useful**

- `OwnerAccountAddress` on retained quotes is taken from `trustedAccount.Address` (mixed case from DB today).
- Locking-cap logic calls `GetRetainedQuotesForAddress(ctx, trustedAccount.Address, ...)`, which may match `owner_account_address` **exactly** in Mongo.
- If one accept stores EIP-55 casing and a later accept recovers lowercase, cap aggregation / per-address queries could treat them as different owners unless everything is normalized downstream.

Setting `Address` to the recovered canonical form at accept time would align runtime state with the signer identity and avoid subtle inconsistencies outside the trusted-account collection.

**Why I didn’t include it here**

- Not part of the review feedback; the mongo `$toLower` lookup fix is sufficient for the reported `TamperedTrustedAccountError`.
- Small behavior change on retained-quote fields — worth a separate discussion/PR if you’d like canonical `OwnerAccountAddress` everywhere.

Happy to add this in a follow-up if you think it’s worth it.

---

### Verification

Unit tests:

```bash
go test ./internal/entities/blockchain/... ./internal/usecases/liquidity_provider/... ./internal/adapters/dataproviders/database/mongo/... -count=1
```

I manually verified authenticated pegin accept on regtest with a mixed-case trusted account + valid signature (no `TamperedTrustedAccountError`).

volodymyrzahanych (approved):

volodymyrzahanych (approved):

PR #965, opened 2026-05-05
FLY-2264 Pegout - Wrong amount message when not enough liquidity in /acceptQuote endpoint

TravellerOnTheRun:
## What
Pegout **`POST /pegout/acceptQuote`** liquidity failure (`NoLiquidityError`) now exposes clear context in the wrapped use-case error **`Args`**: **`quoteValue`**, **`gasFee`**, **`requiredLiquidity`**, and **`missingLiquidity`** (when a follow-up read of available liquidity succeeds). The old single **`amount`** field (always the required total) is removed so it is not mistaken for quote size or shortfall alone.

The use case **preserves the error returned by `HasPegoutLiquidity`** (instead of re-wrapping only the bare sentinel), so provider text such as **`missing … satoshi`** remains in the error string surfaced as **`details.error`** on **HTTP 409**. Errors from **`HasPegoutLiquidity`** that are **not** `NoLiquidityError` are propagated without being misclassified as a liquidity shortage.

## Why
[FLY-2264](https://rsklabs.atlassian.net/browse/FLY-2264): QA reported the accept-quote error was misleading—users could not reliably see **quote value** vs **missing** liquidity. Implementation aligns the server payload with that expectation.

## Type of Change
- [ ] Breaking change 
- [ ] Documentation update
- [ ] Performance improvement
- [x] Test updates
- [ ] Security fix

- [ ] Management UI / API
- [ ] PegIn flow
- [x] PegOut flow
- [ ] Utility scripts
- [ ] Configuration files
- [ ] Metrics and alerting

## Related Issues
[FLY-2264](https://rsklabs.atlassian.net/browse/FLY-2264)

## How to test
- `go test ./internal/usecases/pegout/... -short` (includes `TestAcceptQuoteUseCase_Run_NoLiquidity` and `…_preservesProviderErrorMessage`).
- Optional: `go test -v ./internal/usecases/pegout -run NoLiquidity` and log `err.Error()` to confirm **`Args`** JSON and, in the provider-wrapped case, **`missing … satoshi`** in the same string (this is what the REST layer puts in **`details.error`** for **409** `"not enough liquidity"`).
- Staging / local LPS: reproduce insufficient pegout liquidity (e.g. accept a quote that locks BTC, then accept another) and confirm **409** body: **`message`** unchanged; **`details.error`** contains the new keys and no misleading lone **`amount`**.

## Reviewer Guidelines
- `calculateAndCheckLiquidity` delegates arg shaping to **`newPegoutAcceptNoLiquidityErrorArgs`**; on the no-liquidity path it may call **`AvailablePegoutLiquidity`** once more to fill **`missingLiquidity`**—acceptable tradeoff vs duplicating shortfall math.
- Confirm **backward compatibility** expectations for any client that parsed only **`amount`**.

N/A (API error string / JSON).

## Additional Notes
Branched from **`v2.5.1`**. Public API shape unchanged at top level (**`message`** + **`details.error`** string); only the contents of that string / embedded **`Args`** JSON change for this failure mode.

### DIFF
--- a/internal/usecases/pegout/accept_pegout_quote.go
+++ b/internal/usecases/pegout/accept_pegout_quote.go
@@ -2,6 +2,7 @@ package pegout
 
 import (
 	"context"
+	"errors"
 	"sync"
 
 	"github.com/rsksmart/liquidity-provider-server/internal/entities"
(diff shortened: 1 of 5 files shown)


volodymyrzahanych:
Merge v2.5.1 into fly-2264 and push. This does three useful things at once:

1. Pushes a new commit → re-triggers dependency-review against the new base → annotations clear.
2. Brings the branch up to date with the base it now targets, so the PR diff in the UI is clean and the dev tests on top of current v2.5.1.
3. Standard hygiene — reviewers see what will actually land.

TravellerOnTheRun:
> Merge v2.5.1 into fly-2264 and push. This does three useful things at once:
> 
> 1. Pushes a new commit → re-triggers dependency-review against the new base → annotations clear.
> 2. Brings the branch up to date with the base it now targets, so the PR diff in the UI is clean and the dev tests on top of current v2.5.1.
> 3. Standard hygiene — reviewers see what will actually land.

done!

internal/usecases/pegout/accept_pegout_quote.go
```diff
@@ -197,16 +198,41 @@ func (useCase *AcceptQuoteUseCase) checkLockingCap(ctx context.Context, trustedA
...
-	var err error
 	requiredLiquidity := new(entities.Wei)
-	errorArgs := usecases.NewErrorArgs()
-
 	requiredLiquidity.Add(pegoutQuote.Value, pegoutQuote.GasFee)
-	if err = useCase.pegoutLp.HasPegoutLiquidity(ctx, requiredLiquidity); err != nil {
-		errorArgs["amount"] = requiredLiquidity.String()
-		return nil, usecases.WrapUseCaseErrorArgs(usecases.AcceptPegoutQuoteId, usecases.NoLiquidityError, errorArgs)
+
+	err := useCase.pegoutLp.HasPegoutLiquidity(ctx, requiredLiquidity)
```
volodymyrzahanych, line 204: HasPegoutLiquidity internally already called AvailablePegoutLiquidity.  The added second call to pegoutLp.AvailablePegoutLiquidity does:
- another BTC GetBalance
= another Mongo GetRetainedQuoteByState
…just to recompute missingLiquidity for errorArgs. 

Suggested simpler shape: call AvailablePegoutLiquidity once at the top, compare locally, and skip the redundant HasPegoutLiquidity indirection — or expose the missing amount from HasPegoutLiquidity via a typed error (e.g. NoLiquidityError with fields) so callers don't have to re-derive it.

volodymyrzahanych (commented):
It looks like Pegin has the identical bug — fix not applied symmetrically. Since that is out of current task scope it makes sense to report that and/or create a new task

volodymyrzahanych (commented):
It looks like Pegin has the identical bug — fix not applied symmetrically. Since that is out of current task scope it makes sense to report that and/or create a new task

internal/usecases/pegout/accept_pegout_quote.go
TravellerOnTheRun: ## Summary
- Address the feedback: avoid duplicate `AvailablePegoutLiquidity` (BTC balance + Mongo retained quotes) when accept pegout fails for insufficient liquidity.
- `HasPegoutLiquidity` now returns `InsufficientLiquidityError` with snapshotted `Available`/`Required`; accept pegout uses `errors.As` to build API error args (`missingLiquidity`, etc.) in a single liquidity check.
- Fixes a side effect where `HasPegoutLiquidity` mutated the caller’s `requiredLiquidity` via `Sub` when formatting the error.

volodymyrzahanych (approved):

PR #966, opened 2026-05-07
feat(FLY-2271): detect bridge release_request_rejected in rebalance

volodymyrzahanych:
## What
Parse the Bridge release_request_rejected event from native pegout receipts and mark the rebalance as BridgeTxFailed with the decoded reason (low_amount, caller_contract, fee_above_value, unknown) when present. Previously a successful tx implicitly meant a successful release, so silently rejected releases were recorded as succeeded.

## Why
Currently, the LPS always assumes that the rebalance was accepted, so once it executes the native pegout in the BridgePegoutUseCase, it marks the rebalance as succeeded in the DB with BridgeTxSucceeded unless a explicit error with the TX happens. This ignores the fact that the tx might be successful but the BTC release rejected. The purpose of this tasks is to verify that by checking the logs of the transaction and looking for the event release_request_rejected(address,uint256,int256), if such event is present then the rebalance should be marked as failed as well.

## Type of Change
- [ ] Breaking change 
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test updates
- [ ] Security fix

- [ ] Management UI / API
- [ ] PegIn flow
- [x] PegOut flow
- [ ] Utility scripts
- [ ] Configuration files
- [ ] Metrics and alerting

## Related Issues
[Jira ticket](https://rsklabs.atlassian.net/browse/FLY-2271)

## How to test
That is not so trivial to test in a real environment as there are some guard checks in components
of rootstock ecosystem . I tested locally in the regtest environment with some workarounds including
lps code check bypass to trigger low_amount rejection event in powpeg bridge.
So only unit tests are basically possible.

## Reviewer Guidelines

## Additional Notes


internal/adapters/dataproviders/rootstock/parsing.go
```diff
@@ -6,16 +6,20 @@ import (
...
 	"github.com/ethereum/go-ethereum/common"
 	geth "github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/crypto"
 	"github.com/rsksmart/liquidity-provider-server/internal/adapters/dataproviders/rootstock/bindings"
 	"github.com/rsksmart/liquidity-provider-server/internal/entities"
 	"github.com/rsksmart/liquidity-provider-server/internal/entities/blockchain"
 	"github.com/rsksmart/liquidity-provider-server/internal/entities/quote"
 )
 
+var releaseRequestRejectedTopic = crypto.Keccak256Hash([]byte("release_request_rejected(address,uint256,int256)"))
```
Luisfc68: If you need the abi updated, then open a PR in the LBC repo with the updates to the bridge interface and then sync the bindings in this project, in that way we can guarantee some consistency between what we have in the contracts and here

internal/adapters/dataproviders/rootstock/parsing.go
```diff
@@ -119,3 +123,57 @@ func ParseDepositEvent(receipt blockchain.TransactionReceipt) (blockchain.Parsed
...
+	default:
+		return blockchain.RejectedPegoutReasonUnknown
+	}
+}
+
+func findReleaseRejectedLog(logs []blockchain.TransactionLog, normalizedBridge string) (blockchain.TransactionLog, bool) {
+	for _, log := range logs {
+		if len(log.Topics) > 0 &&
+			bytes.Equal(log.Topics[0][:], releaseRequestRejectedTopic.Bytes()) &&
+			strings.TrimPrefix(strings.ToLower(log.Address), "0x") == normalizedBridge {
```
Luisfc68: why do you remove the 0x outside and here? if you know that the address comes with 0x then why removing it is part of the normalization? Anyway we have a `CompareIgnore0x` for case insensitive comparison ignoring the prexi so you can use that

internal/adapters/dataproviders/rootstock/parsing.go
```diff
@@ -119,3 +123,57 @@ func ParseDepositEvent(receipt blockchain.TransactionReceipt) (blockchain.Parsed
...
+	}
+	if len(log.Data) < 64 {
+		return true, blockchain.RejectedPegoutReasonUnknown
+	}
+	// Decode the trailing int256 reason word using two's complement. big.Int.SetBytes treats
+	// its input as unsigned, so a negative int256 (high bit of the leading byte set) would
+	// otherwise decode as a near-2^256 positive number. If the sign bit is set, subtract 2^256
+	// to recover the signed value. Current Bridge code only emits positive reason codes, but
+	// the param is declared int256 so we decode it as signed for forward compatibility.
+	reasonBytes := log.Data[len(log.Data)-32:]
```
Luisfc68: if you create the binding using abigen like the ones we have in the rest of the project then this fragment is not needed

internal/adapters/dataproviders/rootstock/parsing_test.go
```diff
@@ -285,3 +286,139 @@ func TestParseDepositEvent(t *testing.T) {
 		assert.Empty(t, deposit)
 	})
 }
+
+var releaseRejectedTopic = crypto.Keccak256Hash([]byte("release_request_rejected(address,uint256,int256)"))
```
Luisfc68: this supports the point that is not a good idea to create bindings outside of the place we have for the bindings, then they will be created all around the project and some times even repeated

internal/adapters/dataproviders/rootstock/parsing_test.go
Luisfc68, line 1: I recommend manually taking examples from mainnet for these tests, let me know if you need some

internal/adapters/dataproviders/rootstock/wallet.go
```diff
@@ -155,7 +155,7 @@ func (wallet *RskWalletImpl) buildTransactionReceipt(receipt *geth.Receipt, tx *
 		GasUsed:           new(big.Int).SetUint64(receipt.GasUsed),
 		Value:             txValue,
 		GasPrice:          entities.NewWei(receipt.EffectiveGasPrice.Int64()),
-		Logs:              make([]blockchain.TransactionLog, 0),
+		Logs:              convertReceiptLogs(receipt),
```
Luisfc68: why you opted by using `convertReceiptLogs` instead of `ParseReceipt`?

Luisfc68 (changes requested):

internal/adapters/dataproviders/rootstock/parsing.go
volodymyrzahanych: Declare the event in LBC's IBridge interface, not in LPS — done on LBC branch [fly-2271-event-release_request_rejected](https://github.com/rsksmart/liquidity-bridge-contract), top commit 21c3844 feat: declare release_request_rejected event in IBridge interface. So the contract side is the source of truth.

internal/adapters/dataproviders/rootstock/parsing.go
volodymyrzahanych: Removed the manual two's-complement decode and the now-obsolete "decode signed int256" comment block.
Use bridgeAbi.UnpackIntoInterface(event, releaseRequestRejectedEvent, log.Data) to populate a bindings.IBridgeReleaseRequestRejected. The ABI codec handles int256 signedness, so event.Reason.Int64() is already correct.

internal/adapters/dataproviders/rootstock/parsing_test.go
volodymyrzahanych: Both production and tests source the topic from bindings.IBridgeMetaData.

internal/adapters/dataproviders/rootstock/wallet.go
volodymyrzahanych: good point. Using ParseReceipt(tx, receipt) simplifies that functions significantly

internal/adapters/dataproviders/rootstock/parsing.go
volodymyrzahanych: Good point. I missed utils.CompareIgnore0x

internal/usecases/pegout/bridge_pegout.go
```diff
@@ -86,6 +97,16 @@ func (useCase *BridgePegoutUseCase) Run(ctx context.Context, watchedQuotes ...qu
 	return nil
 }
 
+func (useCase *BridgePegoutUseCase) checkBridgeRejection(receipt blockchain.TransactionReceipt) error {
+	rejected, reason := useCase.releaseRejectionParser(receipt, useCase.contracts.Bridge.GetAddress())
+	if rejected {
+		reasonStr := string(reason)
+		log.Errorf("%s: bridge rejected release for tx %s, reason=%s", usecases.BridgePegoutId, receipt.TransactionHash, reasonStr)
+		return fmt.Errorf("%w: reason=%s", ErrBridgeReleaseRejected, reasonStr)
+	}
```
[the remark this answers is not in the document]
volodymyrzahanych: Copilot's suggestion conflicts with the project's "no defensive validation for impossible scenarios" rule, and the "no-op default" alternative is genuinely dangerous here

internal/adapters/dataproviders/rootstock/parsing.go
```diff
@@ -119,3 +123,57 @@ func ParseDepositEvent(receipt blockchain.TransactionReceipt) (blockchain.Parsed
...
+	r := new(big.Int).SetBytes(reasonBytes)
+	if reasonBytes[0]&0x80 != 0 {
+		r.Sub(r, new(big.Int).Lsh(big.NewInt(1), 256))
+	}
```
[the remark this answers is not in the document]
volodymyrzahanych: Good catch — applied. Copilot is right that Int64() truncation could silently route an out-of-range reason to one of the three known codes (e.g., 2^64 + 1 → 1 → LowAmount), which parseRejectedPegoutReason's default branch wouldn't catch. Added the IsInt64() guard plus a regression test using 2^100.

internal/adapters/dataproviders/rootstock/parsing_test.go
volodymyrzahanych: Add a sub-test using the verbatim release_request_rejected log from
rootstock-testnet tx 0x4d9ee3b5…79f35 (sender 0x62db…d5b0, amount
100000 sat, reason 1) to assert the parser decodes a real Bridge
rejection to RejectedPegoutReasonLowAmount.

internal/adapters/dataproviders/rootstock/parsing.go
```diff
@@ -119,3 +122,55 @@ func ParseDepositEvent(receipt blockchain.TransactionReceipt) (blockchain.Parsed
...
+	case 2:
+		return blockchain.RejectedPegoutReasonCallerContract
+	case 3:
+		return blockchain.RejectedPegoutReasonFeeAboveValue
+	default:
+		return blockchain.RejectedPegoutReasonUnknown
+	}
+}
+
+func findReleaseRejectedLog(logs []blockchain.TransactionLog, bridgeAddress string, topic common.Hash) (blockchain.TransactionLog, bool) {
```
Luisfc68: this can be reused to parse other events I recommend to rename it with generic names for the function and parameters

Luisfc68 (approved):

internal/adapters/dataproviders/rootstock/parsing.go
volodymyrzahanych: rename findReleaseRejectedLog to generic findEventLog

PR #970, opened 2026-05-11
feat: add federation migration script

AndresQuijano:
## What
Adds an optional federation segwit migration step to the local docker-compose environment so the LPS can run against a P2SH-P2WSH (segwit-compatible) federation in regtest.

- Introduces fed-migrator, a new docker-compose service containing a Node.js script (migrateFederationToSegwit.js) that votes a federation change through the RSK bridge using the regtest-authorized keys, mines enough blocks for the new federation to activate, verifies the activation by checking the federation address changed, and writes USE_SEGWIT_FEDERATION=true to the env file.
- Adds MIGRATE_FEDERATION=false flag to sample-config.env; when set to true in the local env file, lps-local.sh runs the migrator before starting LPS and re-sources the env file so LPS picks up the updated flag.
- Updates docker-compose/rskj/rsk.conf to enable the reed800/reed810 hardforks (activating RSKIP-305, which makes the bridge produce P2SH-P2WSH federation addresses) and disables SVP (rskip419 = -1) since BTC spend validation is not required in local dev.

## Why
Make local env more stable

## Type of Change
- [ ] Breaking change 
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test updates
- [ ] Security fix
- [x] Deployment/Infrastructure changes

- [ ] Management UI / API
- [ ] PegIn flow
- [ ] PegOut flow
- [x] Utility scripts
- [ ] Configuration files
- [ ] Metrics and alerting

## Related Issues
[Jira ticket](https://rsklabs.atlassian.net/browse/FLY-2280)

## How to test
Start the full env with and without the new flags CREATE_POWPEG and MIGRATE_FEDERATION. A docker full clean is recommended on every run of the lps-local.sh script


docker-compose/rskj/rsk.conf
```diff
@@ -11,14 +11,6 @@ rpc {
             }
         }
     }
-   modules {
```
Luisfc68, line 14: Pls leave this as is, the reason why it is like this is because we had a security requirement to disable the unused modules in our nodes. Since we don't control that, we provide this configuration in the default values, if you need a different config for the local env I'd suggest create a separated config file for local

Luisfc68: What I mean is that someone uses this as reference then it'd be better if they copy this configuration for the modules, this file is still used for the local config as it is for regtest, but in order to use it as an example thats why I recommend leaving the configs like this

docker-compose/local/fed-migrator/scripts/migrateFederationToSegwit.js
```diff
@@ -0,0 +1,272 @@
...
+        throw new Error(
+            `Federation address did not change after ${BLOCKS_AFTER_COMMIT} blocks.\n` +
+            `  Old address: ${oldFederationAddress}\n` +
+            `  New address: ${newFederationAddress}\n` +
+            'The new federation may not have activated yet — increase BLOCKS_AFTER_COMMIT.'
+        );
+    }
+    console.log('  Federation activation confirmed.');
+
+    // ── Step 9: Update env file ───────────────────────────────────────────────
```
Luisfc68, line 259: why do we need this step? if this step is running is because the env was in true in the first place, right?

docker-compose/local/lps-local.sh
```diff
@@ -64,6 +64,23 @@ fi
...
+    docker compose -f docker-compose.yml -f fed-migrator/docker-compose.fed-migrator.yml --env-file "$ENV_FILE" up --wait
+    EXIT_CODE=$(docker wait fed-migrator)
+    if [ "$EXIT_CODE" != "0" ]; then
+      echo "ERROR: Federation migration failed (exit code $EXIT_CODE)"
+      exit 1
+    fi
+    echo "Federation migrated to segwit"
+    set -a
+    
+    source "$ENV_FILE"
```
Luisfc68, line 80: I think this part is not needed because of my other comment

Luisfc68 (changes requested):

docker-compose/local/fed-migrator/scripts/migrateFederationToSegwit.js
AndresQuijano: No, the condition for this to run is:
CREATE_POWPEG=true
MIGRATE_FEDERATION=true

Once it is migrated to Segwit, it updates this other flag:
USE_SEGWIT_FEDERATION=false -> true

Luisfc68 (approved):

PR #996, opened 2026-05-21
docs(FLY-2308): logging library SPIKE — recommend slog

volodymyrzahanych:
## What
Add SPIKE doc and four runnable demo packages comparing logrus, slog, zap, and zerolog against the Base Logging Standards. Recommendation: adopt log/slog behind a thin LPS wrapper.

## Why

## Type of Change
- [ ] Breaking change 
- [x] Documentation update
- [ ] Performance improvement
- [ ] Test updates
- [ ] Security fix

- [ ] Management UI / API
- [ ] PegIn flow
- [ ] PegOut flow
- [ ] Utility scripts
- [ ] Configuration files
- [ ] Metrics and alerting

## Related Issues
[Jira ticket](https://rsklabs.atlassian.net/browse/FLY-2308)

## How to test
That is just spike task 

## Reviewer Guidelines

## Additional Notes

### DIFF
--- a/docs/spikes/logs.md
+++ b/docs/spikes/logs.md
@@ -1,0 +1,633 @@
+# Logging library comparison and recommendation
+
+## Decision
+
+Adopt `log/slog` behind a thin LPS-owned `internal/logging` package.
+
+The wrapper should expose the native slog key/value API plus a small set of helpers for LPS-specific gaps: initialization, error fields, fatal-equivalent logging, lazy values, trace propagation, PII redaction, and test capture.
+
+Replace `gorilla/handlers.LoggingHandler` with structured request-logging middleware during the migration.
+
+## Demo packages
+
+The four runnable prototypes live under [spike/logging/](../../spike/logging/). Each is its own `main` package with the same per-deliverable file layout (see [Prototype deliverables](#prototype-deliverables) below), so the comparison axes below can be checked against working code:
+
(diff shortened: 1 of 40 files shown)


Luisfc68 (commented):
I think the content of the spike is appropriate and strongly agree with the recommendation of moving to `slog`. Actually, the evolution of the std lib in the latest years might produce this same case in other aspects (e.g. I think with the current routing capabilities we can get rid of gorilla, so perhaps thats the next candidate to analyze). 

Now talking specifically of the implementation, the fact of migrating to `slog` is not enough, to actually enhance the logging of the application the following should be done in my opinion (besides adapting to the standard format the org is chasing):
- Since `slog` is part of the standard library, implement the logger at entity layer, it fits our criteria for the layers boundaries. Dependency injection is always a good practice, but given the fact that this is part of the standard library is not a hard requirement
- Standardize logging, logging inside the use case is always allowed, but perhaps we should leave it for trace level. I think our standard should be log requests and responses between our service and any other component (bitcoind, rskj, and mongodb). We already do it for the database, we need to do the same for the nodes. Similarly, we need to do the same with incoming requests. All of this with info level, so we can save the rest for actual debugging
- The addition of the trace id should be transparent for the rest of the service, it should be some middleware the one capturing it and injecting it into the context. The existing codebase should not be affected. If anything, maybe just passing the context to some new use cases

PR #1001, opened 2026-05-28
FLY-2310: Embed management UI build output

TravellerOnTheRun:
### What

Embed the Vite-built management UI into the LPS binary and serve it under `/management/next/*` (feature-flagged). Serve hashed assets with immutable caching and serve `index.html` with `no-store`.

### Why

Ship the new Management UI without adding runtime filesystem dependencies and without changing the existing `/management` legacy behavior.

### Type of Change

Feature / internal change (static asset serving + build automation + tests).

### Affected part

- `liquidity-provider-server` Go REST entrypoint management routes/handlers
- Build automation (`Makefile`) and docs
- UI build config (`ui/` Vite base path)

### Related Issues

- FLY-2310

### How to test

- `make test`
- `make build`
- With `ENABLE_MANAGEMENT_API=true` and `ENABLE_MANAGEMENT_UI_NEXT=true`, verify `/management/next/` serves the UI and that:
  - `index.html` returns `Cache-Control: no-store`
  - hashed `/management/next/assets/*.js` and `*.css` return correct `Content-Type` and `Cache-Control: public, max-age=31536000, immutable`
- Verify `/management` legacy endpoint behavior remains unchanged.

### Reviewer Guidelines

- Verify route gating is default-off and cannot be accidentally enabled without explicit env var.
- Verify the handler prevents path traversal and only serves embedded assets.
- Confirm cache-control behavior matches the matrix above.
- Confirm `Makefile` enforces UI build before Go build/test and does not require committing `ui/dist`.

### Additional Notes

- Stage 06 verification status is `amber` because `fly-2309` (base ref) has pre-existing failing tests under `make test` (watcher tests); `fly-2310` runs clean.


managementnextui/fs.go
Luisfc68, line 1: why a different directory? is this temporary until the full ui is build? if thats the case its ok, otherwise I think we should just maintain the same package

Luisfc68: Additionally, I recommend to put this at the same level as the current asset dir `internal/adapters/entrypoints/rest/<new dir>`

managementnextui/fs.go
```diff
@@ -0,0 +1,17 @@
+package managementnextui
+
+import (
+	"embed"
+	"io/fs"
+)
+
+//go:embed dist/**
```
Luisfc68, line 8: I'm not sure about the security implications of including the whole content of the directory, pls check if you can be more specific about the files to embed

README.md
```diff
@@ -58,6 +58,22 @@ pnpm run format                  # Prettier write
 
 **CI:** The `ui-check` job runs install, lint, test, and build before Go unit tests and lint jobs. A failing UI gate blocks the rest of the pipeline.
 
+### Embedded Vite build (management UI "next")
+
+The Vite build output (`ui/dist/`) is embedded into the Go binary and served under **`/management/next/*`** when both of these are enabled:
+
+- `ENABLE_MANAGEMENT_API=true` (existing gate)
+- `ENABLE_MANAGEMENT_UI_NEXT=true` (new gate; default off)
```
Luisfc68, line 66: ok so I assumed correctly, is temporary until we deprecated the other one, that is ok. Anyway I still recommend to move the directory to the proper layer

internal/configuration/environment/environment.go
```diff
@@ -215,12 +215,13 @@ type CaptchaEnv struct {
...
 
 type ManagementEnv struct {
-	EnableManagementApi   bool   `env:"ENABLE_MANAGEMENT_API"`
-	SessionAuthKey        string `env:"MANAGEMENT_AUTH_KEY"`
-	SessionEncryptionKey  string `env:"MANAGEMENT_ENCRYPTION_KEY"`
-	SessionTokenAuthKey   string `env:"MANAGEMENT_TOKEN_AUTH_KEY"`
-	UseHttps              bool   `env:"MANAGEMENT_USE_HTTPS"`
-	EnableSecurityHeaders bool   `env:"ENABLE_SECURITY_HEADERS"`
+	EnableManagementApi    bool   `env:"ENABLE_MANAGEMENT_API"`
+	EnableManagementUiNext bool   `env:"ENABLE_MANAGEMENT_UI_NEXT"`
```
Luisfc68, line 219: you need to update the docker compose files accordingly

internal/adapters/entrypoints/rest/routes/management.go
```diff
@@ -8,21 +8,24 @@ import (
...
+	"github.com/rsksmart/liquidity-provider-server/managementnextui"
 )
 
 const (
 	LoginPath  = "/management/login"
 	UiPath     = "/management"
 	StaticPath = "/static/{file}"
 	IconPath   = "/favicon.ico"
+
+	NextUiPath = "/management/next/{path:.*}"
```
Luisfc68, line 20: If we're enabling the next ui via a feature flag then it means that the server is running either that one or the legacy one, not both at the same time. Therefore I recommend not to have a separate path for the new UI

internal/adapters/entrypoints/rest/routes/management.go
```diff
@@ -185,4 +188,14 @@ func GetManagementEndpoints(env environment.Environment, useCaseRegistry registr
 			Handler: handlers.NewSetLiquidityRatioHandler(useCaseRegistry.SetLiquidityRatioUseCase()),
 		},
 	}
+
+	if env.Management.EnableManagementUiNext {
```
Luisfc68, line 192: Similar to my other comment, we don't need to register an alternative path, the path should remain the same and the flag determines which handler is created, I'm also not sure we need another handler, but I haven't got to that part of the PR yet

internal/adapters/entrypoints/rest/routes/management_next_ui_test.go
Luisfc68, line 1: Create this as a test in the other file, otherwise we'll have two test files for the same source files which can lead to confusion

internal/adapters/entrypoints/rest/handlers/management_next_ui.go
Luisfc68, line 1: I have multiple observations on this file, maybe some of them are outside of the PR scope, if thats the case let me know:
- The PR seems to be serving an SPA, as it is not populating the next bundle with any data, what is the rationale to do that instead of using SSR? The main advantage of the SSR approach that we have been using is that we're injecting data directly into the template, this reduces the surface exposed in the API and allows us to keep the UI app simpler in some aspects. I see that the server actually serves another file if the path exists, so pls confirm if the data injection step is outside of the PR scope, in any case, the main way the UI needs to be provided with the data is via injection, save the requests for separate cases with specific criteria
- Continuing with the previous point (in case the SPA assumption is correct), in a scenario where the SPA is actually required then there is no advantage on embedding it into the server, just saving the infrastructure required to expose it, but there wouldn't be any actual difference between putting it here, using a bucket and cloudfront, or running an nginx in the same compose project
- I none of the security headers are present in the handler, these are very important to have as we use them to block multiple attack vectors, pls don't skip them
```
	cspHeader := fmt.Sprintf("default-src 'self'; font-src 'self' data:; style-src 'self' 'sha256-<styles hash>'; object-src 'none'; frame-src 'self'; script-src 'self' 'nonce-%s'; img-src 'self' data:; connect-src 'self';", nonce)
	w.Header().Set("Content-Security-Policy", cspHeader)
	w.Header().Set("Strict-Transport-Security", "max-age=63072000; includeSubDomains; preload")
	w.Header().Set("X-Frame-Options", "DENY")
	w.Header().Set("X-Content-Type-Options", "nosniff")
```

Luisfc68 (changes requested):
I haven't been involved in previous PRs of this epic. Is there any documentation you can point me to to understand how will the integration work specifically? I'm a bit worried about introducing the npm ecosystem into this project with all the supply chain issues it has. Keep in mind that given the close relation between this UI and the server any issue might impact directly into the LP funds. That is the reason why we're ultra conservative with the UI dependencies and the few we have are bundled and stored as part of the server's code (bootstrap and decimal.js). Given the nature of the project I consider the same conservative approach should be maintained, therefore I'd like to review the specific list of dependencies and how is the integration expected to work once is finalized.

managementnextui/fs.go
TravellerOnTheRun: Yes — temporary on purpose. We agreed in daily to keep a separate embed package (managementnextui/) and route (/management/next/) so legacy /management stays unchanged while the epic ships.

TravellerOnTheRun: Agreed long-term home is next to internal/adapters/entrypoints/rest/assets/. For FLY-2310 we kept a top-level embed package to avoid touching legacy asset wiring. Relocate under rest/ when we deprecate the old UI or in a small refactor PR — not blocking this PR.

managementnextui/fs.go
TravellerOnTheRun: Agreed — not required for this PR. Vite output is only index.html plus hashed files under assets/. I will tighten go:embed to explicit paths in a small follow-up PR (separate branch), so this PR stays focused on embed + static serving.

README.md
TravellerOnTheRun: Same as the thread on managementnextui/fs.go: temporary parallel UI until legacy deprecation; directory move to the rest/ layer is planned for cutover/refactor, not required to land the embed + static serve foundation here.

internal/configuration/environment/environment.go
TravellerOnTheRun: Done on fly-2310: ENABLE_MANAGEMENT_UI_NEXT is now passed through in all LPS docker-compose files (same pattern as ENABLE_MANAGEMENT_API), plus docker-compose/local/.env.regtest and sample-config.env document it (default false). No code changes needed in FLY-2311 / FLY-2312 — only rebase those branches onto updated fly-2310 when you pick up the new commit.

internal/adapters/entrypoints/rest/routes/management.go
TravellerOnTheRun: We discussed this in daily: during migration we intentionally run both paths — legacy /management unchanged, new UI at /management/next/ behind ENABLE_MANAGEMENT_UI_NEXT. The flag gates the new route only; it does not mean only one UI exists in the binary yet. Same-path + handler swap is the deprecation/cutover step after the stacked epic PRs land.

internal/adapters/entrypoints/rest/routes/management.go
TravellerOnTheRun: Same rationale as above: separate registration for now avoids changing legacy route wiring. When we flip the default and remove legacy, we will register a single /management (and related static paths) and let the flag choose legacy vs next handler — likely in the same PR that removes the old templates.

internal/adapters/entrypoints/rest/routes/management_next_ui_test.go
TravellerOnTheRun: I would keep the separate test file for now: FLY-2311 (#1003) already extends management_next_ui_test.go with templating/security-header coverage. Folding into management_test.go on 2310 would force a messy rebase on the stacked PRs. Happy to consolidate into one routes test file in a small follow-up after the stack lands (or as part of the last epic PR before cutover).

internal/adapters/entrypoints/rest/handlers/management_next_ui.go
TravellerOnTheRun: Several of these are outside FLY-2310 scope by design; the stack covers them:

**SSR / data injection** — This PR only embeds and serves the Vite build (raw index.html + hashed assets). **FLY-2311** (#1003, stacked on this branch) treats index.html as html/template, injects #initial-data JSON from the same use case as legacy management, and adds CSRF meta. That matches the existing inject-first, API-for-exceptions model.

**Security headers (CSP, HSTS, X-Frame-Options, nosniff)** — Also **FLY-2311**: serveIndex calls the same htmlTemplateSecurityHeaders helper as legacy management.go when ENABLE_SECURITY_HEADERS is set.

**CSRF on mutating API calls** — **FLY-2312** (#1005): React apiFetch reads the meta token and sends X-CSRF-Token on POST/PUT/PATCH/DELETE only.

**Why embed the SPA** — Same deployment model as today (single binary, no runtime ui/dist on disk). The security-sensitive shell is server-rendered in 2311; the client bundle is static assets with immutable cache headers.

**SPA vs SSR for the whole app** — App shell is server-templated; route UI is client-side React under /management/next/ (Vite basename). Epic parent: FLY-2305.

Happy to walk through the stacked PRs if useful — 2310 is intentionally embed + static handler + gates only.

TravellerOnTheRun:
Re: top-level review (npm ecosystem / integration / conservative dependencies):

**Epic context:** Parent [FLY-2305](https://rsklabs.atlassian.net/browse/FLY-2305). This PR is stage 1 only (embed + static serve). Stacked: [#1003 FLY-2311](https://github.com/rsksmart/liquidity-provider-server/pull/1003) (templated shell + security headers + initial-data + CSRF meta), [#1005 FLY-2312](https://github.com/rsksmart/liquidity-provider-server/pull/1005) (React bootstrap + CSRF client interceptor).

**How integration works:**
- `ui/` is built at compile time (`make ui-build` → copy to `managementnextui/dist` → `go:embed`). **No Node/npm at LPS runtime**; only the built artifacts ship in the binary.
- Legacy `/management` is unchanged. New UI is opt-in via `ENABLE_MANAGEMENT_UI_NEXT` at `/management/next/` until cutover.
- Module layout: `docs/ui-folder-structure.md`.

**Dependency posture:** `pnpm-lock.yaml`, CI `ui-check` gate (lint/test/build before Go), minimal runtime deps in `ui/package.json`. We align with the conservative approach (legacy UI only bundles bootstrap + decimal.js in-repo). Happy to do a focused dependency review pass on `ui/package.json` before wider rollout — can share the list in a follow-up comment if helpful.

**Daily alignment:** Separate embed package and `/management/next/` path are intentional during migration (parallel with legacy); same-path/handler swap is a later deprecation PR.

managementnextui/fs.go
Luisfc68: Ok, please put both things as TODO in the code to prevent them from falling through the cracks

managementnextui/fs.go
Luisfc68: Ok, mark it as TODO then

internal/adapters/entrypoints/rest/routes/management.go
Luisfc68: Ok then pls as general comment for all these things that are temporary until the epic is ready pls add the corresponding TODO comments so in the final release there's no trace of having 2 parallel UIs

internal/adapters/entrypoints/rest/routes/management_next_ui_test.go
Luisfc68: Ok, add me as reviewer in the consolidation PR pls

internal/adapters/entrypoints/rest/handlers/management_next_ui.go
Luisfc68: I see, I'll review the PRs for FLY-2311 and FLY-2312 too then, there are some things I want to verify, for example, using current implementation of `htmlTemplateSecurityHeaders` for 2 different UIs might be an error as the styles hash differ between both. The rest sounds logical

Luisfc68:
Before merging pls do the following:
- Add the TODOs mentioned in the comments
- Update the unit tests so the check passes again
- Not for now, but when everything is ready create a full general PR as I think it will be useful to review the epic with the full scope, reviewing stacked PRs can potentially produce errors in the review and/or implementation if not done carefully so to reduce the risk I'd rather take a look in the final PR

PR #1007, opened 2026-06-02
Fix local env

AndresQuijano:
## What
Fixes issues in the local development environment setup scripts and Docker Compose configuration files:
- Renames the `KEYSTORE_FILE` env variable to `WALLET_FILE` across `docker-compose.yml` and `docker-compose.lps-local.yml`
- Adds `--progress plain` flag to `docker compose` commands in `lps-local.sh` for cleaner CI/terminal output
- Adds a workaround in `lps-local.sh` to explicitly start `lps-configurer` if it gets stuck in a `created` state (caused by `lps01` briefly crashing on first boot during BTC wallet rescan)
- Replaces the `lps-configurer` command with a shell wrapper that polls the LPS `/health` endpoint before running the configuration node script
- Changes `lps-configurer`'s `depends_on` condition from `service_healthy` to `service_started` and its restart policy from `"no"` to `on-failure:5`
- Updates `sample-config.env` to enable `CREATE_POWPEG` and `MIGRATE_FEDERATION` by default

## Why
The local environment setup was unreliable due to a race condition: on first boot, `lps01` briefly crashes to perform a BTC wallet rescan, causing `lps-configurer` to miss the healthy window and get permanently stuck in `created` state. The previous `depends_on: service_healthy` approach was too strict and did not account for this transient failure. The `KEYSTORE_FILE` rename aligns the variable name with the actual wallet file concept used elsewhere.

## Type of Change
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test updates
- [ ] Security fix
- [x] Deployment/Infrastructure changes

- [ ] Management UI / API
- [ ] PegIn flow
- [ ] PegOut flow
- [ ] Utility scripts
- [x] Configuration files
- [ ] Metrics and alerting

## Related Issues
[Jira ticket](https://rsklabs.atlassian.net/browse/FLY-2278)

## How to test
1. Run the local environment setup from scratch using `lps-local.sh`
2. Confirm `lps-configurer` starts and completes configuration successfully even on first boot (when `lps01` performs its BTC wallet rescan)
3. Verify that `WALLET_FILE` is correctly passed as an environment variable to the LPS container
4. Check that docker compose output uses plain progress format in both terminal and CI logs

## Additional Notes
The `sample-config.env` changes (`CREATE_POWPEG=true`, `MIGRATE_FEDERATION=true`) update the default values to reflect a more complete local setup out of the box. Adjust as needed for your environment before running.


docker-compose/local/lps-local.sh
```diff
@@ -84,7 +84,14 @@ if [[ "$CREATE_POWPEG" == "true" ]]; then
 fi
 
 ### LPS (always runs) ###
-docker compose -f docker-compose.yml -f lps/docker-compose.lps-local.yml --env-file "$ENV_FILE" up -d --wait
+docker compose --progress plain -f docker-compose.yml -f lps/docker-compose.lps-local.yml --env-file "$ENV_FILE" up -d --wait || true
```
Luisfc68: why the || true? in why case do you want to allow this command to fail?

docker-compose/local/lps/docker-compose.lps-local.yml
```diff
@@ -116,11 +116,18 @@ services:
     volumes:
       - ./lps/configurations:/configurations
       - ./lps/lps_configuration_data:/lps_configuration_data
-    command: ["node", "/configurations/index.js"]
+    command:
```
Luisfc68: Isn't it better to add this as part of the js script? otherwise you have the logic split in both parts, its not wrong to have it there, the script would be making sure the other service is actually running

docker-compose/local/lps/docker-compose.lps-local.yml
```diff
@@ -116,11 +116,18 @@ services:
...
+      - |
+        echo "Waiting for LPS to be healthy..."
+        until wget -q --spider http://lps:8080/health 2>/dev/null; do sleep 3; done
+        echo "LPS is healthy, running configuration..."
+        node /configurations/index.js
     depends_on:
       lps:
-        condition: service_healthy
-    restart: "no"
+        condition: service_started
```
Luisfc68, line 122: why would you want to start the configurer before the LPS is healthy?

Luisfc68 (commented):
the specifics were discussed in a call

docker-compose/local/lps/docker-compose.lps-local.yml
AndresQuijano: The LPS is going to fail first time and then restart. If this check (which is a gate, not a pool) is executed at that moment, the configuration process will be never executed. Instead, the LPS at this point will always be started.
The pooling, until it is healthy, is going to be executed in the waitForHealth() function.

docker-compose/local/lps/docker-compose.lps-local.yml
AndresQuijano: Yes, moving it there

docker-compose/local/lps/docker-compose.lps-local.yml
Luisfc68: making the script wait for health is ok, but the LPS should not fail, if it shuts down because of the rescan that is suppose to end the process with 0, pls check that because if its not having that behavior we need to update it

docker-compose/local/lps-local.sh
```diff
@@ -84,7 +84,17 @@ if [[ "$CREATE_POWPEG" == "true" ]]; then
 fi
 
 ### LPS (always runs) ###
-docker compose -f docker-compose.yml -f lps/docker-compose.lps-local.yml --env-file "$ENV_FILE" up -d --wait
+# lps01 may briefly crash on first boot during a BTC
```
Luisfc68, line 87: pls remove this comment as this behavior is not normal, if the process finished it should not crash

Luisfc68 (approved):

Luisfc68:
I left the PR approved but make sure all the checks pass before merge pls

PR #1008, opened 2026-06-02
Feat/fly 2331

Hakob23:
## What
Add 6 structured `logrus.WithFields(...)` log statements to the validation phase of `AcceptQuoteUseCase.Run` in `internal/usecases/pegin/accept_pegin_quote.go`:

- `Debug` "Accepting pegin quote" at `Run` entry (`quoteHash`, `hasSignature`)
- `Warn` "Accept pegin rejected: contract paused" on `CheckPauseState` error
- `Error` "Accept pegin: failed to load quote" on `GetQuote` error
- `Warn` "Accept pegin rejected: quote not found" when `GetQuote` returns `nil`
- `Warn` "Accept pegin rejected: quote expired" on `IsExpired()` (includes `expireTime`)
- `Warn` "Accept pegin rejected: locking cap exceeded" in `checkLockingCap` (includes `trustedAccount`, `currentLocked`, `lockingCap`, `newQuoteValue`)

Threaded `quoteHash` through `handleTrustedAccountSignature` and `checkLockingCap` so the locking-cap log can include it.

Added matching `test.AssertLogContains` assertions across the existing validation-phase tests and a new dedicated `TestAcceptQuoteUseCase_Run_GetQuoteError` test for the GetQuote-error branch.

## Why
Accepting a pegin quote is a money-binding decision — the LP signs a quote and reserves liquidity. Today, when an accept fails or behaves unexpectedly in production we only have the HTTP-level error and no log trail tying the failure back to the `quoteHash`, signature presence, or rejection reason. These logs make every validation-phase rejection greppable/aggregatable in production so on-call and audits can reconstruct what happened.

## Type of Change
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [x] Test updates
- [ ] Security fix
- [x] Observability / logging (no behavior change)

- [ ] Management UI / API
- [x] PegIn flow
- [ ] PegOut flow
- [ ] Utility scripts
- [ ] Configuration files
- [ ] Metrics and alerting

## Related Issues
[Jira ticket](<insert URL>)

## How to test
1. `go test ./internal/usecases/pegin/...` — all tests pass, including the 5 updated validation tests and the new `TestAcceptQuoteUseCase_Run_GetQuoteError`.
2. `go vet ./internal/usecases/pegin/...` — clean.
3. Manual smoke (optional): hit the accept endpoint with bad inputs and confirm each rejection path emits its log line:
   - Pause the LBC pegin contract → expect `Accept pegin rejected: contract paused`.
   - Accept a non-existent `quoteHash` → expect `Accept pegin rejected: quote not found`.
   - Accept an expired quote → expect `Accept pegin rejected: quote expired` (with `expireTime` field).
   - Accept a signed quote for a trusted account whose active locked liquidity + new quote exceeds the configured `RbtcLockingCap` → expect `Accept pegin rejected: locking cap exceeded` (with `currentLocked`, `lockingCap`, `newQuoteValue`).
   - Force a repo failure on `GetQuote` → expect `Accept pegin: failed to load quote` at `Error` level.

## Reviewer Guidelines
- Behavior is unchanged — only logging was added. Error wrapping, return values, mutex handling, and event publishing are all identical.
- All 6 messages are constant strings (no `Sprintf`) so they can be grepped/aggregated. Variable data lives in structured fields.
- `quoteHash` is present on every log.
- The new `Run` entry log is at `Debug` and is intentionally not asserted in tests because logrus defaults to `InfoLevel` (asserting it would require mutating the global log level, which leaks across tests). All other 5 logs are asserted via `test.AssertLogContains`.
- `handleTrustedAccountSignature` and `checkLockingCap` are package-private helpers — the new `quoteHash` parameter has no external callers.
- The new dedicated `TestAcceptQuoteUseCase_Run_GetQuoteError` was split out of the parameterized `TestAcceptQuoteUseCase_Run_ErrorHandling` so a single failure mode can have a focused log assertion; the parameterized test still covers all the other error setups.
- Out of scope per ticket: success-path / persist / publish logs and PII redaction.

N/A — backend-only logging change.

## Additional Notes
Repo today has only one structured-logging example (`internal/adapters/entrypoints/watcher/monitoring/asset_metrics_watcher.go:64`); this PR follows that style. There is no `log.Ctx(ctx)` helper in the tree, so plain `log.WithFields(...)` + `log.WithError(err)` is used as the ticket allows.


Hakob23:
The refactoring changes are limited to supporting the new logs without changing behavior: `getQuote` and the inline pause check in `Run` were merged into `loadValidQuote`, which returns a `*quote.PeginQuote` and centralizes the validation-phase rejections that now get logged; `quoteHash` was threaded through `validateTrustedAccountIfFound `and `checkLockingCap` so the locking-cap rejection can include it in structured fields; `Run `was cleaned up with `:=` declarations and renaming the idempotent `GetRetainedQuote `result to existing; and `buildRetainedQuote `now takes a pointer to avoid copying the quote through the call chain. Error handling, mutex usage, trusted-account flow, and return values are unchanged relative to v2.6.0 — this is structural cleanup to make logging straightforward, plus `SafeLogStr ` in `common.go ` for CodeQL-safe log fields.

internal/usecases/pegin/accept_pegin_quote.go
```diff
@@ -200,6 +212,14 @@ func (useCase *AcceptQuoteUseCase) checkLockingCap(ctx context.Context, trustedA
 
 	// Check if the sum exceeds the locking cap
 	if totalWithNewQuote.Cmp(trustedAccount.RbtcLockingCap) > 0 {
+		newQuoteValue := new(entities.Wei).Add(peginQuote.Value, peginQuote.GasFee)
```
AndresQuijano: newQuoteValue is computed here as peginQuote.Value + peginQuote.GasFee, but that same sum was already computed two lines above as part of totalWithNewQuote (lines 210–211). Probably it can be extracted before the addition so it can be reused:

```
newQuoteValue := new(entities.Wei).Add(peginQuote.Value, peginQuote.GasFee)
totalWithNewQuote := new(entities.Wei).Add(totalLocked, newQuoteValue)
if totalWithNewQuote.Cmp(trustedAccount.RbtcLockingCap) > 0 {
    log.WithFields(log.Fields{
        "newQuoteValue": newQuoteValue.String(),
        ...
    })
```

internal/usecases/pegin/accept_pegin_quote.go
```diff
@@ -108,16 +106,50 @@ func (useCase *AcceptQuoteUseCase) Run(ctx context.Context, quoteHash, signature
 	}, nil
 }
 
+func (useCase *AcceptQuoteUseCase) loadValidQuote(
+	ctx context.Context, quoteHash string, logger *log.Entry,
```
AndresQuijano: Just a matter of style and consistency:  
loadValidQuote receives a *log.Entry from Run, but checkLockingCap builds its own entry locally from quoteHash. Since loadValidQuote already has quoteHash as a parameter, it can build the same logger itself.
Why the strategy is different here?

internal/usecases/common.go
```diff
@@ -174,6 +174,14 @@ func WrapUseCaseError(useCase UseCaseId, err error) error {
 	return WrapUseCaseErrorArgs(useCase, err, make(ErrorArgs, 0))
 }
 
+// SafeLogStr strips CR/LF from a user-controlled string before it is included
+// in a structured log field, preventing log forging (CWE-117).
+func SafeLogStr(s string) string {
```
AndresQuijano, line 179: It's a small and trivial function but I'd add small and trivial unit test as in the other functions of this file.

internal/usecases/pegin/accept_pegin_quote.go
AndresQuijano, line 202: Why did you decided not to put some logs on this error path?

AndresQuijano (changes requested):

AndresQuijano:
Since we are modifying the way we log messages, I'd recommend keeping the literals in a common place so we can use them as reference in the code and tests.

internal/usecases/common.go
Hakob23: https://github.com/rsksmart/liquidity-provider-server/pull/1008/commits/3f819a96509d6bd6b41b6ddb725c38fc4c3aad73

internal/usecases/pegin/accept_pegin_quote.go
Hakob23: The task was initially about validation-phase rejections, but it's good to have a log here as well. Added with the test

PR #1013, opened 2026-06-10
fix: remove fatal on rescan

AndresQuijano:
## What

Replace two `log.Fatal` + `os.Exit(1)` paths in the BTC wallet initialization flow with a sentinel error and a clean `os.Exit(0)`. When the Bitcoin wallet is either mid-rescan or has just triggered one, the server now logs an informational message and exits with code 0 instead of crashing with code 1.

## Why

The BTC wallet rescan is a known, expected part of the LPS lifecycle on every first boot with a new wallet. Using `log.Fatal` (which calls `os.Exit(1)`) was semantically wrong:

- It misrepresents the situation to operators — fatal signals something unrecoverable, not a scheduled background process.
- It misrepresents the situation to process supervisors (Docker, systemd, Kubernetes) — exit code 1 looks like a crash, consuming failure budgets and triggering false alerts.
- On mainnet, a rescan can take hours. A supervisor with a limited restart budget (e.g. `on-failure:3`) exhausts that budget long before the rescan finishes, leaving LPS permanently stopped.

## Type of Change

- [x] Utility scripts
- [x] Configuration files

## Related Issues

[FLY-2380](https://rsklabs.atlassian.net/browse/FLY-2380)

## How to test

1. Run a fresh local environment
2. On first boot, observe that `lps01` logs `"Bitcoin wallet rescan in progress. The server will start once the rescan completes. Exiting cleanly."` and exits with code 0 (not 1).
3. Docker automatically restarts `lps01` (via `unless-stopped` policy) until the rescan completes and the server boots normally.
4. Confirm `lps-configurer` eventually exits 0 after LPS becomes healthy at `/health`.
5. On a subsequent boot (rescan already done), confirm LPS starts normally with no scanning-related log lines.

## Additional Notes

`cmd/application/**` was added to the codecov ignore list to fix the `codecov/patch` check. The 9 new lines in `newBitcoinRegistry` (`cmd/application/lps/application.go`) are inherently untestable — the function calls `os.Exit(0)` and `log.Fatal`, which terminate the process and cannot be exercised in unit tests without injecting an exit function (a separate refactor outside this ticket's scope). The composition root (`cmd/application/`) has no existing tests for the same reason, and the project has always excluded this layer from coverage expectations. `cmd/utils/**` (which does have tests) is unaffected by this exclusion.


Luisfc68 (changes requested):
The code in general is fine but I think there is no need to remove the path from codecov as the function can still be tested with multiple approaches e.g. bubble the error until main or inject the exit function in the application class

